PR #7, opened 2019-04-18
Internal CRDs

scothis:
Decompose Application and Function into a single common type Plan. Plan then decomposes into a BuildPlan and a RunPlan as needed.

The behavior enables a few things:
- Functions no longer decompose to Applications, each namespace is unique
- BuildPlan decomposes to Knative Build types, but can be implemented with other systems (like Tekton)
- RunPlan decomposes to Knative Serving types
- Build and Run are independent, and optional. An image can be built without being run, an existing image can run without a custom build. This enables more evolved CD patterns to exist above Function and Application.
- Knative Services are no longer used, Configurations are used instead. This enables continuous deployment of updated images, even while splitting routing in a blue/green environment. Advanced routing is not exposed by this PR.


pkg/reconciler/v1alpha1/buildplan/buildplan.go
```diff
@@ -0,0 +1,385 @@
...
+
+func (c *Reconciler) reconcileBuild(ctx context.Context, buildplan *planv1alpha1.BuildPlan, build *buildv1alpha1.Build) (*buildv1alpha1.Build, error) {
+	logger := logging.FromContext(ctx)
+	desiredBuild, err := resources.MakeBuild(buildplan)
+	if err != nil {
+		return nil, err
+	}
+
+	if build != nil && desiredBuild == nil {
+		// TODO drop a existing build
```
sbawaska: Log something until we get to implement the TODO?

trisberg:
should we reject Applications that don't specify a `template` for the build part? They seem to never get built but there are no obvious errors either.

scothis:
@trisberg Applications in this PR doesn't specify a template, the controller injects them with the cf-cnb template. That template isn't part of any PR yet, but it is available here https://github.com/scothis/builder/blob/application-builder/cf-cnb-clusterbuildtemplate.yaml

trisberg:
applying the `sample/application.yaml` creates an Application resource with an empty template:

```
spec:
  build:
    cacheSize: 8Gi
    source:
      git:
        revision: master
        url: https://github.com/cloudfoundry-samples/cf-sample-app-nodejs
    template: ""
  image: gcr.io/cf-sandbox-sandrews/sample
  run:
    resources: {}
```

it never creates the build with or without a `cf-cnb` clusterbuildtemplate

scothis:
Moving forward with #8 instead

PR #8, opened 2019-04-23
Build all the things

scothis:
A rethinking of #7 that focuses on builds without concern for the container runtime. A followup PR will address runtime concerns.

Splits build and run-time artifacts for Application and Functions into a new API group build.projectriff.io. InternalBuild is not intended for direct public use.

Reconciliation:
- ApplicationBuild -> InternalBuild
- FunctionBuild -> InternalBuild
- InternalBuild -> PVC (cache, if needed)
- InternalBuild -> Build (Knative)

The digest of the built image is set as `.status.latestImage`.

These build resources are embeddable within a Knative Service or Configuration and will participate within the existing Knative serving build-run lifecycle.


markfisher:
I feel like InternalBuild is leaking out an implementation detail as a CR instance. Would it be possible to structure shared code to handle that (mapping to Knative Build for now) instead while maintaining the modularity that enables future changes to be isolated to a well-defined scope?

PR #13, opened 2019-05-03
Stream processing CRDs

scothis:


config/300-stream.yaml
```diff
@@ -0,0 +1,39 @@
+# Copyright 2019 The original author or authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
```
ericbottard: Do we really want licences on yaml files?
And if so, I think this should use https

scothis: maybe the bot will send a PR...

PR #15, opened 2019-05-14
Reconcile credentials to riff-build ServiceAccount

scothis:
Credentials are Secrets with a label marking them as a riff Credential.
The credentails should be bound to the riff-build service account in the
namespace the secrets exist. When a credential is deleted, the service
account should no longer contain a reference to that secret.

Other non-credential secrets are not impacting durring these operations.
An annotation is added to the service account to indicate which secrets
are credentials.


scothis:
@trisberg did we want to do something with pull secrets as well?

PR #17, opened 2019-05-20
Resolve Processor image from Function resource

scothis:
- Processor.Spec.Function -> Process.Spec.FunctionRef
- tracks changes to referenced Function
- set's resolved lastest image as Processor.Status.FunctionImage
- adds webhook validation for Stream and Processor CRDs
  - normalizes ObjectMeta validation


markfisher:
LGTM

PR #21, opened 2019-05-24
Apply default image prefix to apps and funcs

scothis:
The image may be either exactly `_` or start with `_/`. For `_` the
default image prefix is combined with the applicaiton/function name. For
`_/` the underscore is replaced with the default image prefix. In both
cases, the resolved image is stashed in Status.TargetImage.

The reconcile will lookup the riff-build config map in the same
namespace as the application/function when the image starts with `_`. If
the config map does not exist or does not have a default-image-prefix
value, the reconciler will error.


scothis:
@trisberg combined with projectriff/riff#1293 you'll now see errors like

```
$ riff function list
NAME   LATEST IMAGE   ARTIFACT   HANDLER   INVOKER   STATUS                      AGE
bad    <empty>        <empty>    <empty>   <empty>   DefaultImagePrefixMissing   23m
```

scothis:
Merging to unstick riff

PR #26, opened 2019-05-30
Define RBAC ClusterRoles for riff resources

### LINKED ISSUE #11: define cluster roles for end users to interact with the riff system resources
For example, one role may provide write access to riff resources, and another role provides read-only access.

scothis:
There are two ClusterRoles defined:
- `projectriff` read/write access to all projectriff runtime resources
- `projectriff-readonly` same as projectriff, but without mutations

Cluster operators may bind these roles to service accounts, users and/or
groups to either the whole cluster or a specific namespace via a
ClusterRoleBinding or RoleBinding respectively.

These roles are also aggregated to the default `admin`, `edit` and
`view` ClusterRoles.

See https://kubernetes.io/docs/reference/access-authn-authz/rbac/#rolebinding-and-clusterrolebinding

Fixes #11


scothis:
@trisberg these ClusterRoles are intended for use by operators to grant users access to riff resources. Our controllers will define their own roles and bindings with access to all of the resources they need.

PR #28, opened 2019-05-31
Expose content types of output streams

fbiville:


pkg/reconciler/v1alpha1/processor/resources/deployment.go
```diff
@@ -91,3 +75,36 @@ func MakeDeployment(proc *streamv1alpha1.Processor) (*appsv1.Deployment, error)
...
+			Value: processor.Name,
+		},
+		{
+			Name:  "FUNCTION",
+			Value: "localhost:8080",
+		},
+	}, exposeOutputContentTypes("OUTPUT_CONTENT_", processor.Status.OutputContentTypes)...)
+}
+
+func exposeOutputContentTypes(keyPrefix string, outputContentTypes []string) []corev1.EnvVar {
```
fbiville: If you wonder why I go to great length to expose each content type separately, that's because it seems there is no reliable way to list content-type in a single envvar (no delimiter seems safe): https://www.w3.org/Protocols/rfc1341/4_Content-Type.html

pkg/apis/stream/v1alpha1/processor_types.go
```diff
@@ -42,10 +42,11 @@ type ProcessorSpec struct {
 type ProcessorStatus struct {
 	duckv1alpha1.Status `json:",inline"`
 
-	InputAddresses  []string `json:"inputAddresses,omitempty"`
-	OutputAddresses []string `json:"outputAddresses,omitempty"`
-	DeploymentName  string   `json:"deploymentName,omitempty"`
-	FunctionImage   string   `json:"functionImage,omitempty"`
+	InputAddresses     []string `json:"inputAddresses,omitempty"`
+	OutputAddresses    []string `json:"outputAddresses,omitempty"`
+	OutputContentTypes []string `json:"outputAddresses,omitempty"`
```
ericbottard: Copy/paste is the root of all evil :) see json tag

pkg/reconciler/v1alpha1/processor/resources/deployment.go
```diff
@@ -52,29 +53,12 @@ func MakeDeployment(proc *streamv1alpha1.Processor) (*appsv1.Deployment, error)
...
-								{
-									Name:  "GROUP",
-									Value: proc.Name,
-								},
-								{
-									Name:  "FUNCTION",
-									Value: "localhost:8080",
-								},
-							},
+							Image: "ericbottard/processor:grpc-be83df8", // TODO: update for output content-type support
```
fbiville: depends on https://github.com/ericbottard/streaming-processor-poc/pull/2/

pkg/reconciler/v1alpha1/processor/resources/deployment.go
```diff
@@ -91,3 +75,36 @@ func MakeDeployment(proc *streamv1alpha1.Processor) (*appsv1.Deployment, error)
...
+		},
+		{
+			Name:  "GROUP",
+			Value: processor.Name,
+		},
+		{
+			Name:  "FUNCTION",
+			Value: "localhost:8080",
+		},
+	}, exposeOutputContentTypes("OUTPUT_CONTENT_", processor.Status.OutputContentTypes)...)
```
ericbottard: No need to parameterize the prefix, in the same way that INPUTS, OUTPUTS, etc is hardcoded (or could become a const) here

fbiville: it's just to co-locate them all, it serves no other purposes.

pkg/apis/stream/v1alpha1/stream_types.go
```diff
@@ -37,6 +37,7 @@ type Stream struct {
 
 type StreamSpec struct {
 	Provider string `json:"provider"`
+	ContentType string `json:"content-type"`
```
scothis:
```suggestion
	ContentType string `json:"contentType"`
```

pkg/apis/stream/v1alpha1/stream_validation.go
```diff
@@ -31,16 +31,20 @@ func (s *Stream) Validate(ctx context.Context) *apis.FieldError {
 	return errs
 }
 
-func (ss *StreamSpec) Validate(ctx context.Context) *apis.FieldError {
-	if equality.Semantic.DeepEqual(ss, &StreamSpec{}) {
+func (s *StreamSpec) Validate(ctx context.Context) *apis.FieldError {
```
scothis: please revert

pkg/apis/stream/v1alpha1/stream_validation.go
```diff
@@ -31,16 +31,20 @@ func (s *Stream) Validate(ctx context.Context) *apis.FieldError {
...
 	}
 
 	errs := &apis.FieldError{}
 
-	if ss.Provider == "" {
+	if s.Provider == "" {
 		errs = errs.Also(apis.ErrMissingField("provider"))
 	}
 
+	if s.ContentType == "" {
```
scothis: is the content type actually required?

pkg/apis/stream/v1alpha1/stream_types.go
scothis: Is "ContentType" the proper name for what we want here?

pkg/apis/stream/v1alpha1/stream_validation.go
```diff
@@ -31,16 +31,20 @@ func (s *Stream) Validate(ctx context.Context) *apis.FieldError {
...
+		errs = errs.Also(apis.ErrMissingField("content-type"))
```
scothis:
```suggestion
		errs = errs.Also(apis.ErrMissingField("contentType"))
```

pkg/reconciler/v1alpha1/processor/resources/deployment.go
scothis: perhaps a higher level config format should be defined

pkg/apis/stream/v1alpha1/stream_validation.go
fbiville: it's reported by Goland. It suggests all receivers should be named consistently, I do not see the harm in that.

pkg/apis/stream/v1alpha1/stream_validation.go
fbiville: yes, otherwise the invoker (which gets the start frame with the output content types computed by the processor) won't be able to find appropriate converters.
I don't think there can be a sensible default here.

pkg/apis/stream/v1alpha1/stream_validation.go
scothis: Goland is wrong, the behavior was consistent before this change. `*Stream` != `*StreamSpec`

pkg/apis/stream/v1alpha1/stream_validation.go
ericbottard: I _think_ we could try to leave this blank and have the invokers support the lack of content type.
`application/octet-stream` is also a good, equivalent, default

pkg/reconciler/v1alpha1/processor/resources/deployment.go
ericbottard: Could do, although this is entirely private to riff and never exposed to the user

pkg/apis/stream/v1alpha1/processor_types.go
fbiville: oopsie :)

pkg/apis/stream/v1alpha1/stream_validation.go
fbiville: yep, or the inconsistency is somewhere else, probably misread it

ericbottard:
I'd be in favor of merging this PR as is for now and refine if necessary

ericbottard:
Although CI gives 
```
ERROR: /home/vsts/work/1/s/gopath/src/github.com/projectriff/system is out of date. Please run ./hack/update-codegen.sh
``` 
but @fbiville  managed to use the value of the field. Maybe you did not re-generate after the json tag update?

pkg/reconciler/v1alpha1/processor/resources/deployment.go
scothis: > Could do, although this is entirely private to riff and never exposed to the user

All the more reason we can pass a blob of json or some other format guaranteed to never have encoding issues.

pkg/apis/stream/v1alpha1/stream_defaults.go
```diff
@@ -22,6 +22,6 @@ func (s *Stream) SetDefaults(ctx context.Context) {
 	s.Spec.SetDefaults(ctx)
 }
 
-func (spec *StreamSpec) SetDefaults(ctx context.Context) {
-	// TODO
+func (ss *StreamSpec) SetDefaults(ctx context.Context) {
+	ss.ContentType = "application/octet-stream"
```
scothis: this will overwrite any existing value

fbiville: yep, just noticed it in practice.
Ambiguous docs FTW :(
```
// Defaultable defines an interface for setting the defaults for the
// uninitialized fields of this instance.
type Defaultable interface {
	SetDefaults(context.Context)
}
```

pkg/apis/stream/v1alpha1/stream_validation.go
fbiville: done now

fbiville:
> I'd be in favor of merging this PR as is for now and refine if necessary

Yep, still needs https://github.com/ericbottard/streaming-processor-poc/pull/2 first.

pkg/reconciler/v1alpha1/processor/resources/deployment.go
fbiville: done now

ericbottard:
Merged as 2564114

PR #29, opened 2019-06-04
Define an interface to make ready checks easier

scothis:
Introduce Object and Status interfaces for riff objects. Object can get
the Status and Status can get the ready condition. These interfaces will
make writing clients that need to watch for the ready state to go green
or red easier.


pkg/apis/object.go
```diff
@@ -0,0 +1,38 @@
...
+package apis
+
+import (
+	duckv1alpha1 "github.com/knative/pkg/apis/duck/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+type Object interface {
+	runtime.Object
```
ericbottard: Seems good overall, but why require those 3 additional "super-interfaces"? Why not just ask for the capability to return a Status?

scothis: done

scothis:
Merged as 10ae5f6984a2a04ec9bc9a0a2db84c782640780c

PR #31, opened 2019-06-14
Support unauthenticated builds

scothis:
Allow Application and Function builds without first creating a
Credential. Previously without a Credentials the riff-build
ServiceAccount would not be created, causing the build pods to fail. Now
we create the ServiceAccount when a Credential, Application or Function
are created.

This also means that the riff-build ServiceAccount may be deleted. If
the SA is not needed, it will not be recreated.


trisberg:
If I run a function build without credentials I now see this

```
$ riff functions list
NAME     LATEST IMAGE   ARTIFACT    HANDLER   INVOKER   STATUS      AGE
square   <empty>        square.js   <empty>   <empty>   not-Ready   3m49s

$ kubectl get functions
NAME     READY   REASON
square   False   
```

Maybe we could add the failure above somewhere?

```
  - lastTransitionTime: "2019-06-14T17:33:05Z"
    message: 'build step "build-step-export" exited with code 1 (image: "docker-pullable://projectriff/builder@sha256:afc3ccc57046034030dd662e2cc30427a09c944a93a73cd1a64bea825e39d90c");
      for logs run: kubectl -n default logs square-function-pod-f6b893 -c build-step-export'
    status: "False"
    type: BuildSucceeded
```

scothis:
Merged because this PR doesn't change how errors are propagated or reported.

@trisberg can you open an issue with the full resource yaml and what you expect to see in the listing

PR #32, opened 2019-06-26
Update Knative to v0.7

scothis:
Switch from Knative Service v1alpha1 to v1beta1 types.

Also, fail CI on test failures 🤦‍♂️

Depends on https://github.com/projectriff/cnab-riff/pull/30


scothis:
Need to revert the serving api to v1alpha1 due to https://github.com/knative/serving/issues/4533

PR #37, opened 2019-07-24
Spike: drop Knative Serving

scothis:
Replace the Knative Configuration with a Deployment and the Route with a Service.

This spike is missing:
- autoscaling
- ingress

Usage:

```sh
$ riff handler create upper --function upper
$ kubectl port-forward service/upper-handler 8080:80
Forwarding from 127.0.0.1:8080 -> 8080
Forwarding from [::1]:8080 -> 8080
```

```sh
$ curl localhost:8080 -d "hello" -H "Content-Type: text/plain" -w'\n'
HELLO
```


scothis:
Closing, we'll introduce a k8s native runtime that sits along side the Knative runtime rather than replacing it

PR #39, opened 2019-07-30
Add Container CRD

scothis:
Define a build-less build CRD as `containers.build.projectriff.io`. That
will resolve an image to a digest using the same mechanism as
Application and Function.

There are now three developer abstractions:
- Function - build with function buildpacks
- Application - build with application buildpacks
- Container - use a pre built image

Tools that can consume riff build types can point to a Container
resource to pick up new images with the same interface as Application
and Function.

Allow Handlers to consume Containers


markfisher:
should we consider (not as part of this PR) renaming the "build" token in the api group domain to "workload"?

scothis:
> should we consider (not as part of this PR) renaming the "build" token in the api group domain to "workload"?

Interesting idea. It would help Container fit better since it's a buildless build. "workload" has a runtime connotation in my head, but it makes sense in this context.

pkg/apis/request/v1alpha1/handler_validation.go
```diff
@@ -79,6 +79,12 @@ func (b *Build) Validate(ctx context.Context) *apis.FieldError {
 		unused = append(unused, "applicationRef")
 	}
 
+	if b.ContainerRef != "" {
+		used = append(used, "containerRef")
+	} else {
+		unused = append(unused, "contaienrRef")
```
markfisher: typo: contaienr

pkg/reconciler/v1alpha1/container/container.go
```diff
@@ -0,0 +1,225 @@
+/*
+Copyright 2018 The Knative Authors
```
markfisher: 2019

sample/container.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: build.projectriff.io/v1alpha1
+kind: Container
+metadata:
+  name: upper
+spec:
+  image: _
```
markfisher: I realize it's not new to this, but looking at in context, this feels unintuitive if someone isn't aware of the defaulting, and it seems almost like it shouldn't be required and that should simply be the default if left empty in the spec. Also the comment in the container_types code states:

>  May contain a leading underscore to have the default image prefix applied.

That only refers to "leading" and "prefix", not the defaulting of the image name itself.

scothis: I updated the comment and apply `_` to empty images in the defaulter. For these resource, I'd still favor being explicit about the image. The previous values were hard coded images in my gcr bucket that nobody else could pull, so `_` is a huge improvement.

PR #40, opened 2019-07-31
Add Knative runtime

scothis:
Use an Adapter resource to bind the latest image from a build to a
Knative Service or Configuration. The updated images is set using the
v1beta1 types; v1alpha1 types will be converted up.

The adapter does not create or control any other resource. Any other
controller of the target Knative resource must tolerate the image being
mutated.

```yaml
apiVersion: knative.projectriff.io/v1alpha1
kind: Adapter
metadata:
  name: my-adapter
spec:
  build:
    functionRef: my-riff-function
    # applicationRef: my-riff-application
  target:
    serviceRef: my-knative-service
    # configurationRef: my-knative-configuration
```


sbawaska:
If the idea is to decouple function/application/container from runtime which could be knative/streaming etc, would the Adapter be better off outside the `knative.projectriff.io/v1alpha1` api group? The Adapter could then map the build to the runtime.

scothis:
> would the Adapter be better off outside the knative.projectriff.io/v1alpha1 api group? 

This is the adapter for Knative resources. The adapter implementation needs deep knowledge of the target runtime in order to successfully update the image (Knative Serving's API surface is a bit of a kludge in the transition from v1alpha1 to v1beta1). Other runtimes can have their own adapter that maps to resources of that type.

The goal is to be able to eventually carve off the Knative runtime into a separate install so that users who do not want Knative are not forced to install it.

PR #43, opened 2019-08-02
Rename Container CRD to Image

scothis:


scothis:
Keeping the Container name for now

PR #49, opened 2019-08-12
Don't wait for knative runtime informers unless using Knative runtime

scothis:
Waiting for them causes the controller to hang.


cmd/controller/main.go
```diff
@@ -129,6 +129,11 @@ func main() {
 			knativeRuntime = true
 		}
 	}
+	if knativeRuntime {
+		logger.Info("Starting with Knative runtime")
+	} else {
+		logger.Warn("Starting without Knative runtime")
```
sbawaska: Why a warning? Should this instead be: "starting with core runtime"?

scothis: The core runtime is always installed.

It's a warning to make it more distinguished in the logs

PR #51, opened 2019-08-13
Refine deployment ready condition

scothis:
Avoid a trap condition where the deployment can get stuck as not ready


scothis:
Merged as 12aabaa17d416b18d9473e4e60b1a021456af93f

PR #55, opened 2019-08-15
Update AUTHORS

trisberg:
We should either update or remove this file


scothis:
This repo is forked from knative serving. So all of those entities do have a copyright claim to part of the code.

trisberg:
Maybe we should put that in a NOTICE file since they are not active in riff but we include their work

PR #56, opened 2019-08-15
fix handler reference

sbawaska:


scothis:
This is intention. The function handler predated the Handler resource. This is actually the primary driver for renaming the Handler resource

PR #60, opened 2019-08-22
Create a keda scaler upon riff processor creation.

ericbottard:


pkg/reconciler/v1alpha1/streamingprocessor/streamingprocessor.go
```diff
@@ -57,11 +61,14 @@ const (
 type Reconciler struct {
 	*reconciler.Base
 
+	kedaClientSet kedaclientset.Interface
```
scothis: let's treat the Keda clientset the same as other clientsets by making it a property on the controller. This will aid in testing as we can provide stub clientsets in a consistent way that record actions that can be asserted.

See
- https://github.com/projectriff/system/blob/master/cmd/controller/main.go#L133-L136
- https://github.com/projectriff/system/blob/master/pkg/reconciler/v1alpha1/streamingprocessor/streamingprocessor_test.go#L66-L70
- https://github.com/projectriff/system/blob/master/pkg/reconciler/v1alpha1/testing/listers.go

cmd/controller/main.go
```diff
@@ -266,6 +278,7 @@ func main() {
 	kubeInformerFactory.Start(stopCh)
 	projectriffInformerFactory.Start(stopCh)
 	knbuildInformerFactory.Start(stopCh)
+	kedaInformerFactory.Start(stopCh)
```
scothis: we don't want to require that Keda be installed for all runtimes, much like we don't want Knative Serving to be required for all runtimes.

pkg/reconciler/v1alpha1/streamingprocessor/resources/deployment.go
```diff
@@ -44,7 +45,7 @@ func MakeDeployment(proc *streamingv1alpha1.Processor) (*appsv1.Deployment, erro
 			Labels:          makeLabels(proc),
 		},
 		Spec: appsv1.DeploymentSpec{
-			Replicas: &one,
+			Replicas: &zero,
```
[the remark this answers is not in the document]
ericbottard: I think this is (and already was) taken care of by this piece of logic, isn't it:

```golang
func (c *Reconciler) reconcileDeployment(ctx context.Context, processor *streamingv1alpha1.Processor, deployment *appsv1.Deployment) (*appsv1.Deployment, error) {
	logger := logging.FromContext(ctx)
	desiredDeployment, err := resources.MakeDeployment(processor)
	if err != nil {
		return nil, err
	}

	// Preserve replicas as is it likely set by an autoscaler
	desiredDeployment.Spec.Replicas = deployment.Spec.Replicas
```

cmd/controller/main.go
ericbottard: Implemented via f05f64b

pkg/reconciler/v1alpha1/streamingprocessor/streamingprocessor.go
ericbottard: Implemented via 2515deb

pkg/reconciler/v1alpha1/streamingprocessor/streamingprocessor.go
```diff
@@ -19,9 +19,12 @@ package streamingprocessor
 import (
 	"context"
 	"fmt"
+	kedav1alpha1 "github.com/kedacore/keda/pkg/apis/keda/v1alpha1"
```
scothis: move down below

sample/keda.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: keda.k8s.io/v1alpha1
```
scothis: do we need this? It's created by the reconciler

pkg/reconciler/v1alpha1/streamingprocessor/streamingprocessor.go
```diff
@@ -330,8 +368,50 @@ func (c *Reconciler) createDeployment(processor *streamingv1alpha1.Processor) (*
 	}
 	return c.KubeClientSet.AppsV1().Deployments(processor.Namespace).Create(deployment)
 }
-
 func deploymentSemanticEquals(desiredDeployment, deployment *appsv1.Deployment) bool {
 	return equality.Semantic.DeepEqual(desiredDeployment.Spec, deployment.Spec) &&
 		equality.Semantic.DeepEqual(desiredDeployment.ObjectMeta.Labels, deployment.ObjectMeta.Labels)
 }
+
+func scaledObjectSemanticEquals(desiredScaledObject, scaledObject *kedav1alpha1.ScaledObject) bool {
```
scothis: the SemanticEquals method is traditionally after the reconcile method for the type

ericbottard:
Merged as squashed 52c54d4

PR #63, opened 2019-09-10
Define Provider CRD

fbiville:


fbiville:
Note: I haven't been able to fully run the hack scripts.

scothis:
@fbiville this PR will need to be updated to be built upon  kubebuilder. For the most part the stream_types.go can copy on over, but take a look at processor_types.go to get a sense for how things work. The biggest change is the client is no longer statically generated.

PR #77, opened 2019-09-12
Create riff-build service account attaching credentials

scothis:
- create the riff-build ServiceAccount when there is an Application or
  Function build, or there is a credential Secret
- attach credential Secrets to the riff-build ServiceAccount


pkg/controllers/build/credentials_controller.go
```diff
@@ -0,0 +1,226 @@
...
+	if secretNames.Len() != 0 {
+		return true, nil
+	}
+	var applications buildv1alpha1.ApplicationList
+	if err := r.List(ctx, &applications, client.InNamespace(namespace)); err != nil {
+		return false, err
+	} else if len(applications.Items) != 0 {
+		return true, nil
+	}
+	var functions buildv1alpha1.FunctionList
```
sbawaska: should we check for containers as well?

scothis: Containers don't produce build pods, so they don't need a service account

PR #83, opened 2019-09-16
Stop reconciling resources that are being deleted

scothis:
A resource may exist while it is being deleted (finalizers). We should
not continue to reconcile the resource and assume that it is going away
very shortly.


ericbottard:
Merged as 749cb96

Issue #67, opened 2019-09-10
implement controllers for build


sbawaska:

scothis:
Done in #70

PR #87, opened 2019-09-17
Switch to kpack

scothis:
from https://github.com/pivotal/kpack/tree/master/pkg/apis/build/v1alpha1

Refs #67 projectriff/riff#1348


pkg/apis/thirdparty/kpack/build/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,41 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package v1 contains API Schema definitions for the kpack v1alpha1 API group
+//
+// This API group is a forked subset of https://github.com/pivotal/kpack/tree/master/pkg/apis/build/v1alpha1
```
sbawaska: I know this was a problem for knative since the dependencies there were very large. Is that true with kpack as well?

ericbottard: Yeah, did the same for keda, but not sure I understand the rationale, either wrt size or version:
- size: only code actually used is linked
- versions: if our dependency depends on a different version of say a k8s API, that itself is versioned in a different package (assuming compatibility semantics post-alpha, agreed).

scothis: The primary reason for this is to avoid version hell latter on, not just with system, but also with anything that consumes system (like the cli, or other tools). There were many times where upgrading Knative was difficult not because of API changes, but because of the dependency graph. In once case, we ended up with a version deadlock because Knative Serving and Knative Build had different dependencies.

By importing the types into our project we have a clean dependency graph that is centrally managed and can be tested. Since we're only including the structs that represent the public API, these are unlikely to change in significant ways absent an API version bump (of course alphas will be alphas).

When we introduce a dependency like Keda or kpack, we're not linking against it's runtime, we're consuming it's CRD types. If these projects (or us) were not using go, we'd have to take the same action of creating structs for the types. I'm avoiding the artificial shortcut that incidentally adds massive dependency graph complexity.

Like it or not, the Kubernetes go APIs are not perfectly backwards compatible. They do make breaking changes. We should pick the version of the Kubernetes APIs we want to consume.

scothis: If you need third party validation of this approach, consider [Knative and Istio](https://github.com/knative/pkg/tree/master/apis/istio/v1alpha3)

scothis:
This PR now also includes controllers for Applications and Functions to be backed by kpack

pkg/controllers/build/application_controller.go
```diff
@@ -61,38 +59,36 @@ func (r *ApplicationReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error)
...
 	application.SetDefaults(ctx)
 	application.Status.InitializeConditions()
 
-	result, err := r.reconcile(ctx, log, application)
+	result, err := r.reconcile(ctx, log, &application)
 
 	// check if status has changed before updating, unless requeued
 	if !result.Requeue && !equality.Semantic.DeepEqual(application.Status, originalApplication.Status) {
 		// update status
-		log.Info("updating application status", "application", application.Name,
```
sbawaska: why not have additional logging may be at debug level?

pkg/controllers/build/application_controller.go
```diff
@@ -101,86 +97,30 @@ func (r *ApplicationReconciler) reconcile(ctx context.Context, log logr.Logger,
 		}
 		return ctrl.Result{}, err
 	}
-	log.V(1).Info("resolved target image", "application", application.Name, "image", targetImage)
```
sbawaska: keep debug level logging?

pkg/controllers/build/application_controller.go
```diff
@@ -196,30 +136,135 @@ func (r *ApplicationReconciler) resolveTargetImage(ctx context.Context, applicat
...
+	// delete image if no longer needed
+	if desiredImage == nil {
+		if err := r.Delete(ctx, &actualImage); err != nil {
+			log.Error(err, "unable to delete kpack Image for Application", "image", actualImage)
+			return nil, err
+		}
+		return nil, nil
+	}
+
+	// create iamge if it doesn't exist
```
sbawaska: typo

pkg/controllers/build/application_controller.go
scothis: We need to address logging. I tend to do better by thinking about a component of the system in a cross cutting fashion. (Address logging for all controllers at once) rather than address is piecemeal with gap in one controller but not another.

There are other similar concerns, like tests that I'd like to defer until we get basic functionality in place and then address comprehensively.

scothis:
We now have a builder that is compatible with Buildpack lifecycle 0.4, but are blocked by kpacks incompatibility with unauthenticated registries.

Ref https://github.com/pivotal/kpack/issues/143

PR #89, opened 2019-09-18
publish the runtimes

sbawaska:
still a work in progress.
Will have to limit to master before merging.


.github/workflows/pull_request.yaml
```diff
@@ -0,0 +1,44 @@
+name: CI
```
scothis: it should be possible to combine this with the master (previously ci) workflow adding a condition to the publish job to restrict it to pushes to the master branch.

See https://help.github.com/en/articles/workflow-syntax-for-github-actions#jobsjob_idstepsif, but for a job instead of a step (it exists on AZP)

.github/workflows/master.yaml
```diff
@@ -0,0 +1,157 @@
...
+            cut -d '"' -f 4 |\
+            xargs curl -O -L
+          mv kustomize_*_${opsys}_amd64 kustomize
+          chmod u+x kustomize
+          sudo mv kustomize /usr/local/bin/
+      - name: Package
+        run: make manager package
+      - name: gcloud auth
+        run: |
+          gcloud config set project cf-spring-pfs-eng
```
scothis: is gcloud finally on the machine by default? 🕺

.github/workflows/master.yaml
```diff
@@ -0,0 +1,157 @@
...
+          gcloud config set compute/region us-central1
+          gcloud config set compute/zone us-central1-a
+          gcloud config set disable_prompts True
+          gcloud auth activate-service-account --key-file <(echo $GCLOUD_CLIENT_SECRET)
+          gcloud auth configure-docker
+        env:
+          GCLOUD_CLIENT_SECRET: ${{ secrets.GCLOUD_CLIENT_SECRET }}
+      - name: publish images
+        run: |
+          readonly version=$(cat VERSION)
```
scothis: I'd tend to move this into a shell script rather than keeping it inside yaml, but do what you think is right.

.github/workflows/pull_request.yaml
sbawaska: although undocumented for job,`if` works! will make the change.

.github/workflows/master.yaml
sbawaska: yes!

.github/workflows/master.yaml
sbawaska: done!

PR #92, opened 2019-09-19
Make config/riff-*.yaml useable

scothis:
The kustomized component configs are now directly applyable. The images
use the commited version as a tag. The publish script will retag the
snapshot images to match the version tag.

Also further normalizing the publish scripts.


scothis:
Closing in favor of #93

PR #93, opened 2019-09-20
Replace Dockerfile with ko to build images

scothis:
🎶 reunited, and it feels so good 🎶

To deploy a development build:

    ko apply -f config/riff-${component}.yaml


Makefile
```diff
@@ -16,48 +13,27 @@ GOBIN=$(shell go env GOBIN)
...
-docker-push: docker-build ## Push the docker image
-	docker push ${IMG}
-
 .PHONY: test
 test: generate fmt vet manifests ## Run tests
 	go test ./... -coverprofile cover.out
 
-.PHONY: package
-package: generate fmt vet manifests ## Package for distribution
+.PHONY: compile
```
ericbottard: What's the point of this target? It publishes images (with forced -L = local) and their coordinates are lost.

I need a way to publish images to a possibly non local, configurable registry (ie use KO_DOCKER_REPO) and apply the resulting yaml

Makefile
```diff
@@ -123,6 +99,15 @@ else
 GOIMPORTS=$(shell which goimports)
 endif
 
+# find or download ko, download ko if necessary
+ko:
+ifeq (, $(shell which ko))
+	GO111MODULE=off go get github.com/google/ko/cmd/ko
```
ericbottard: Can we force a preferred version?

config/build/manager/manager.yaml
```diff
@@ -23,11 +23,9 @@ spec:
         control-plane: controller-manager
     spec:
       containers:
-      - command:
-        - /manager
-        args:
+      - args:
```
ericbottard: Why did the args disappear in all 4 places?

config/build/manager/manager.yaml
```diff
@@ -23,11 +23,9 @@ spec:
...
         - --enable-leader-election
-        image: controller:latest
+        image: github.com/projectriff/system/cmd/managers/build
         name: manager
```
ericbottard: Also, would be nice to move name and/or image as the top level field, making the list-of-maps easier to read

config/build/manager/manager.yaml
ericbottard: nvm, realized `ko` now rewrites the entrypoint for us

Makefile
scothis: we could, but it's not clear that we should. The only tag in the repo is from march and that's before go module support was added. The docs for `ko` suggest installing the latest.

Makefile
scothis: > I need a way to publish images to a possibly non local, configurable registry

That's exactly why the command shouldn't be a make target. Using `ko` directly is more expressive and powerful. What you're looking for is:

```
ko apply -f config/riff-${component}.yaml
```

The intent of the compile target is to make sure that the containers can be built, we don't do anything with those containers. It's only really used as part of the CI test job, I'd be ok dropping the target from the makefile and only running `ko resolve` from the CI workflow.

config/build/manager/manager.yaml
scothis: ko sets the entry point for the container. Keeping the definition in the deployment creates a craslhoop because `/manager` isn't an executable in the container.

config/build/manager/manager.yaml
scothis: I'm indifferent, but would accept a contribution.

.github/workflows/stage.sh
```diff
@@ -9,26 +9,19 @@ readonly git_sha=$(git rev-parse HEAD)
...
+stageComponent() {
   local component=$1
 
-  ( cd config/$component/manager && kustomize edit set image controller=$IMG_REPO/$component-controller:$slug )
-  COMPONENT=$component make docker-push
+  echo ""
+  echo "# Stage riff System: ${component}"
+  echo ""
+  ko resolve -P -t ${slug} -f config/riff-${component}.yaml > bin/riff-${component}.yaml
+  gsutil cp -a public-read bin/riff-build.yaml gs://projectriff/riff-system/snapshots/riff-build-${slug}.yaml
```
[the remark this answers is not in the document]
scothis: good catch

scothis:
> personally, I would have preferred dockerfile to ko's black magic

`ko` is significantly faster than the Dockerfile build, especially on rebuilds, orders of magnitude. But it's real value is that it handle source and config together in a consistent way. The kustomize/Dockerfile based builds approach this is replacing was a kludge. Here's what we had to do previously (at a high level):

1. cd into a very specific directory and edit the kustomize config to set the image
2. re-render the kustomized output to pickup the changed config
3. run docker build with the output set to the same image we set in the kustomize config
4. push that docker image to a registry
5. resolve the image to a digest, so that k8s will see that the image has changed when updating the config, nothing worse then wondering why your code changes don't work only to later discover k8s was running a previous image (we didn't do this previously, but should)
6. use kubectl to apply the config to a cluster
7. repeat steps 1-6 for every other component

Here's the experience with ko:
1. `ko apply -f config/riff-build.yaml`

That's it. Done.

Want to build and run all of the components? `ko apply -f config/`. 💥

scothis:
> we are going further away from kubebuilder's scaffolding, which might make upgrading to a new version difficult

The value of the initial scaffolding is inconsequential to the ongoing maintenance and developer productivity. Kubebuilder does a really good job at maintaining code in an elegant way. The initial scaffolding is just one opinion for project structure. The point of scaffolding is to accelerate the bootstrapping of a new project, not to lock you in. Codegen tools that lock you into a specific structure always fail. I'm glad kubebuilder learned that lesson.

ericbottard:
Merged as c370610

PR #100, opened 2019-09-25
Aggregate ClusterRoles to view, edit and admin

scothis:
Give users access to riff resources if they have the `view`, `edit` or
`admin` ClusterRoles bound.

`view` has read-only access while `edit` and `admin` have full access.


sbawaska:
Should `create` and `delete` be only be available to `admin`s? i.e. leave them out for the `edit` role?

config/build/rbac/aggregated_roles.yaml
```diff
@@ -0,0 +1,40 @@
...
+kind: ClusterRole
+metadata:
+  name: role
+  labels:
+    # Add these permissions to the "admin" and "edit" default roles.
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+rules:
+- apiGroups:
+  - build.knative.dev
```
sbawaska: do we want to do this for the apiGroup `build.projectriff.io` as well?

scothis: that's an epic typo

scothis:
> Should create and delete be only be available to admins? i.e. leave them out for the edit role?

An editors is a normal user who needs to get work done, typically in the context of a single namespace. An admin is an operator who typically setups up environments and policies for other people to then work within.

PR #101, opened 2019-09-26
Use namespace in Gateway address.

ericbottard:
When addressed from outside the same namespace (eg from riff-system),
the gateway needs a fully resolvable DNS name.


ericbottard:
This has been merged via 5d9d6c5ab34301e31bec9c8095a4444c5b76f6d0

PR #106, opened 2019-10-04
Track dependent resources before fetching them

scothis:
This way we can be notified of its creation if it doesn't exist.


scothis:
merged as ef364c106453fc4a05527020989efc1d5fa5c86b

PR #110, opened 2019-10-04
Reflect the latest resolved image on Deployers

scothis:
This means that the Deployer's status will update when the target build
has a new latest image. The Adapter resource already follows this
pattern.


pkg/controllers/core/deployer_controller.go
```diff
@@ -284,6 +284,9 @@ func (r *DeployerReconciler) constructDeploymentForDeployer(deployer *corev1alph
 			},
 		},
 	}
+	if deployment.Spec.Template.Spec.Containers[0].Image == "" {
```
sbawaska: The deployment will not pick up the latest image if one was set previously. This is a change from the previous behavior. remove the if condition?

scothis: We previously used this field as a holder for the resolved image, but it was never persisted back to the api server. If the container image is defined, then it wins.

Previously we validated that either an image or build ref was specified, but not both. We should bring that back.

scothis: one thing we could do is surface the value of the PodSpec container as lastestImage when there is no build.

sbawaska:
Thanks! Ship it!

PR #111, opened 2019-10-04
Log update diffs for controller resources and status

scothis:


scothis:
@sbawaska if this pattern looks good, I'll apply it to the runtimes as well

PR #116, opened 2019-10-08
Avoid child name collisions and deadlocked creates

scothis:
Avoids two distinct issues:
1. name collisions of child resources
2. deadlocked create when status update fails after creating fixed name
   child resource

The name collision is common if a knative and core deployer of the same
name is created. The create deadlock is rare, but has caused a number of
FATS failures.

These issues are resolved by using generated names rather than fixed
names, and the reconciler no longer depends on the child resource name
set in the status. The reconciler will instead list all child resources
with the target resource as the owner.

The listing should return with zero or one item. If zero, a new resource
is created, if one, that resources is reconciled. If there happens to
be more than one item, they are all deleted and then a new clean
resource is created.


ericbottard:
Merged as d451872 and previous commit

PR #119, opened 2019-10-09
adding needed fields to scaled object

sbawaska:
config validation resulted in error since these were missing


pkg/controllers/streaming/processor_controller.go
```diff
@@ -264,6 +264,8 @@ func (r *ProcessorReconciler) constructScaledObjectForProcessor(processor *strea
 			PollingInterval: &one,
 			CooldownPeriod:  &thirty,
 			Triggers:        triggers(processor),
+			MinReplicaCount: &one,
```
ericbottard: Doesn't that prevent scale-to-zero?

sbawaska: fixed!

PR #146, opened 2019-10-21
Pin processor image

scothis:
This will help ensure that stable builds remain stable


ericbottard:
Merged as 88e80de using new processor naming scheme

PR #149, opened 2019-10-24
Webhook support

scothis:


scothis:
This PR adds webhooks to apply defaults and validate our custom resources (similar to the behavior in 0.4). The general pattern is prescribed by kubebuilder. That means that riff now has a dependency on cert-manager 0.10. The install instructions will need to be updated once this change is pulled into the charts repository.

I don't like having a hard dependency on a common tool that just made a backwards incompatible change, but it's good enough for now.

ericbottard:
Merged as 8f59cf1 and a handful of parent commits

PR #152, opened 2019-10-28
Use fixed version of kustomize.

ericbottard:
This is what I used in http-gateway. Seems like overkill
to be "too-smart", in particular now that kustomize will
be available as part of kubectl.


scothis:
I updated to use the [install instructions](https://github.com/kubernetes-sigs/kustomize/blob/master/docs/INSTALL.md#quickly-curl-the-latest-binary) from the kustomize site. 

`kubectl kustomize` doesn't work with our namespace patch right now. Maybe with kubectl 1.17.

PR #159, opened 2019-11-04
Add input/outputNames to processor

### LINKED ISSUE #57: Add named parameter/returnValue suport
This will let invokers:

- support parameter injection by name (applicable at least to Node)
- know the function arity

ericbottard:
Fixes #57


config/riff-streaming.yaml
```diff
@@ -175,18 +175,37 @@ spec:
           description: ProcessorSpec defines the desired state of Processor
           properties:
             functionRef:
+              description: FunctionRef references a function in this namespace.
               type: string
+            inputNames:
+              description: InputNames is a list of argument names, to be used by languages
```
fbiville:
```suggestion
              description: InputNames is an ordered list of argument names, to be used by languages
```

fbiville: applies below as well.

fbiville: or removed all mentions of order, possibly leading to chaos 😨

PR #165, opened 2019-11-06
Use a PodSpec to set advanced properties for the function container

### LINKED ISSUE #155: Allow processors to bring their own image
We shouldn't require that processors reference a function. It must be possible to provide a pre-existing image and run it as a processor. This will be critical for users who have their own CD workflows or have acceptance tests for build images that need to pass before they go into production.

scothis:
Fields like:
- resources requests/limits
- env vars
- volume mounts
- container image

can all be set on the PodSpec, which is then merged with other fields
and containers within the reconciler.

The FunctionRef is no longer a required field, as the image can be
defined directly on the container.

The defaulter ensures there is always at least one container defined.
The validator ensures that only allowed fields are set and that either a
FunctionRef or a container Image is defined, but not both.

Resolves #155


pkg/apis/streaming/v1alpha1/processor_types.go
```diff
@@ -46,6 +47,9 @@ type ProcessorSpec struct {
 	Inputs []StreamBinding `json:"inputs"`
 	// Outputs references an ordered list of streams to bind as outputs
 	Outputs []StreamBinding `json:"outputs"`
+
```
ericbottard: Is `// +optional` still a thing or does it have no consequences whatsoever?
(This comment should go on line 44, but I can't put it there.)

pkg/apis/streaming/v1alpha1/processor_defaults.go
```diff
@@ -44,4 +47,14 @@ func (s *ProcessorSpec) Default() {
...
 		}
 	}
+
+	if s.Template == nil {
+		s.Template = &corev1.PodSpec{}
+	}
+	if len(s.Template.Containers) == 0 {
+		s.Template.Containers = append(s.Template.Containers, corev1.Container{})
+	}
+	if s.Template.Containers[0].Name == "" {
```
ericbottard: Should we always overwrite?

pkg/apis/streaming/v1alpha1/processor_types.go
scothis: it's used when generating the openapi schema on the crd. Added to ProcessorSpec, but should probably be reviewed globally.

pkg/apis/streaming/v1alpha1/processor_defaults.go
scothis: we could, but why?

scothis: it would be better as a validation constraint in that case

ericbottard: Agreed. I think the container names are _implicitly_ part of the contract, so should be enforced.

PR #166, opened 2019-11-06
Lookup Processor images from Container "builds"

scothis:
In addition to a FunctionRef, the Processor can now also resolve an
image from a Container via a ContainerRef. Both the FunctionRef and
ContainerRef fields are colocated in a Build struct with is on the
ProcessorSpec.

This is a breaking change for users of `.spec.functionRef`, the field is
now at `.spec.build.functionRef`. Which is inline with how the core/knative
Deployers behave

Refs #155


ericbottard:
LGTM, feel free to merge alongside https://github.com/projectriff/cli/pull/124

PR #170, opened 2019-11-07
Fail reconciliation of Processor if Stream not ready

### LINKED ISSUE #117: Fail streaming processor reconciliation when stream not ready
@sbawaska and I observed that, sometimes, created streams do not expose their address at all. 
Yet, the processor gets reconciled with bogus stream addresses (i.e. the result of an empty gateway and empty topic separated by slash, in other words `"/"`) and ends up failing way later than it could.

The processor should not reconcile unless all streams' statuses are fully initialized.

ericbottard:
Fixes #117

### DIFF
--- a/pkg/controllers/streaming/processor_controller.go
+++ b/pkg/controllers/streaming/processor_controller.go
@@ -474,16 +474,20 @@ func (r *ProcessorReconciler) resolveStreams(ctx context.Context, processorCoord
 			Name:      binding.Stream,
 		}
 		var stream streamingv1alpha1.Stream
+		if err := r.Client.Get(ctx, streamNSName, &stream); err != nil {
+			return nil, nil, err
+		} else if !stream.Status.IsReady() {
+			return nil, nil, fmt.Errorf("stream %q is not Ready", streamNSName)
+		}
+
+		addresses = append(addresses, stream.Status.Address.String())
+		contentTypes = append(contentTypes, stream.Spec.ContentType)
+
 		// track stream for new coordinates
 		r.Tracker.Track(
 			tracker.NewKey(stream.GetGroupVersionKind(), streamNSName),
 			processorCoordinates,
 		)
-		if err := r.Client.Get(ctx, streamNSName, &stream); err != nil {
-			return nil, nil, err
-		}
-		addresses = append(addresses, stream.Status.Address.String())
-		contentTypes = append(contentTypes, stream.Spec.ContentType)
 	}
 	return addresses, contentTypes, nil
 }
--- a/pkg/controllers/streaming/processor_controller.go
+++ b/pkg/controllers/streaming/processor_controller.go
@@ -173,45 +173,53 @@ func (r *ProcessorReconciler) reconcile(ctx context.Context, logger logr.Logger,
 		return ctrl.Result{}, fmt.Errorf("could not resolve an image")
 	}
 
+	outcome := outcome{}
+	forceTearDown := false
 	// Resolve input addresses
 	inputAddresses, _, err := r.resolveStreams(ctx, processorNSName, processor.Spec.Inputs)
-	if err != nil {
-		return ctrl.Result{Requeue: true}, err
+	if err == nil {
+		processor.Status.InputAddresses = inputAddresses
+	} else {
+		outcome = outcome.coalesce(requeueing(err))
+		forceTearDown = true
 	}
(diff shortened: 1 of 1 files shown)


pkg/controllers/streaming/processor_controller.go
```diff
@@ -455,16 +455,20 @@ func (r *ProcessorReconciler) resolveStreams(ctx context.Context, processorCoord
...
 			return nil, nil, err
+		} else if !stream.Status.IsReady() {
+			return nil, nil, fmt.Errorf("stream %q is not Ready", streamNSName)
 		}
+
 		addresses = append(addresses, stream.Status.Address.String())
 		contentTypes = append(contentTypes, stream.Spec.ContentType)
+
+		// track stream for new coordinates
+		r.Tracker.Track(
```
scothis: Why did you move where the stream is tracked? I would think you'd still want to track the stream, especially when it's not yet ready.

pkg/controllers/streaming/processor_controller.go
```diff
@@ -455,16 +455,20 @@ func (r *ProcessorReconciler) resolveStreams(ctx context.Context, processorCoord
...
 		var stream streamingv1alpha1.Stream
-		// track stream for new coordinates
-		r.Tracker.Track(
-			tracker.NewKey(stream.GetGroupVersionKind(), streamNSName),
-			processorCoordinates,
-		)
 		if err := r.Client.Get(ctx, streamNSName, &stream); err != nil {
 			return nil, nil, err
+		} else if !stream.Status.IsReady() {
+			return nil, nil, fmt.Errorf("stream %q is not Ready", streamNSName)
```
scothis: What should happen if the processor was previously created when a stream becomes unready?

pkg/controllers/streaming/processor_controller.go
ericbottard: Does it matter? I thought the re-consideration was triggered by the requeue of the processor.
If it *does* matter, shouldn't we `Track` all streams as a first step, regardless of errors then? (here the loop exits at the first encountered error)

pkg/controllers/streaming/processor_controller.go
ericbottard: Good question, although I don't think we can witness that in practice now, unless we're really trying to.
I would argue that the deployments should be removed, wdyt?

pkg/controllers/streaming/processor_controller.go
scothis: > Does it matter? I thought the re-consideration was triggered by the requeue of the processor.

Consider if the Processor doesn't change, but a Stream that wasn't ready becomes ready. The Processor will not be reconciled until the normal resync (~10 hours later). If you're not tracking the Stream, then the Processor will not react to the Stream becoming ready. 

> If it does matter, shouldn't we `Track` all streams as a first step, regardless of errors then?

Actually, no. While we could, it's not necessary. The reason is because we need all Streams to be ready. Once the Stream that failed becomes ready the Processor will be reconciled again, and it will make further progress.

pkg/controllers/streaming/processor_controller.go
scothis: I'm fine with removing/not creating the processor when any stream is not ready. We may need to refine that behavior down the road, but we can cross the bridge when we get to it.

pkg/controllers/streaming/processor_controller.go
ericbottard: > The Processor will not be reconciled until the normal resync (~10 hours later)

Isn't that for the case for `Requeue: false`? Here, if this function returns a non nil error, we return `Requeue: true`

Anyway, I'm fine with this block moving up again.

pkg/controllers/streaming/processor_controller.go
ericbottard: Please TAL @ 97bf58c

scothis:
I tried a different approach in #188. The key different is that it doesn't try to artificially continue the reconciliation, but instead let it complete naturally. Instead of forcing the deletion of the Deployment and ScaledObject when a Stream is not ready, they are preserved, the max replicas for the ScaledObject is set to zero when the streams are not ready.

This approach should be beneficial as the code paths are more straight forward, and there is less churn creating and removing resource.

ericbottard:
Went with #188

PR #173, opened 2019-11-07
Consume helperless FATS

scothis:


sbawaska:
hmm, cannot comment on lines that should be removed, but looks like references to helper.sh should be removed.

PR #174, opened 2019-11-07
Rename liiklus to gateway

scothis:
Use the component's role in the system as its name instead of the
implementation technology. Particularly as part of the public API, we
should favor stable names that won't need to change even if we adopt a
different technology, or that tech is rebranded.


ericbottard:
Merged as 8c95a16

Issue #142, opened 2019-10-17
Ingress support in the core runtime


scothis:
One of the key main benefits of the Knative runtime over the Core runtime is that it provides ingress routing to workloads. We should fill that gap.

Design principals:
- only use built-in (core) kubernetes types, like Ingress
- be unopinionated about the ingress controller implementation (Istio vs Contour vs Nginx vs ...)
  - protect users from need to know about the particular ingress implementation
- allow workloads to be exposed by or protected from ingress (public vs private workloads)
- be ready to pickup whatever Ingress.next resource is in core k8s

Other things to keep in mind:
- dns mapping to workloads and namespaces
- ssl certificate provisioning (possible vs manual vs automatic)

andrew-su:
As of right now, ingress is still in beta: `networking.k8s.io/v1beta1` and `extensions/v1beta1`.

andrew-su:
Ingresses I tried the implementation on:

✅ Nginx
✅ Gloo
✅ Contour

jldec:
@scothis @andrew-su 
Do you think we can push this ingress capability one step further to make it usable for demos in environments like docker and minikube? E.g. could we document the steps to install and configure something like the nginx ingress controller?

shashwathi:
@jldec : sure I can look into this next week.

trisberg:
Minikube is super easy
```
minikube addons enable ingress
```
The nginx ingress server is exposed on port 80 on the `minikube ip` address

trisberg:
Just adding some notes, installed ingress using the stable Helm chart:

```
helm install --name nginx-ingress --namespace nginx-ingress stable/nginx-ingress --wait

riff core deployer create upper --image trisberg/upper

ingress=$(kubectl get svc/nginx-ingress-controller -n nginx-ingress -ojsonpath='{.status.loadBalancer.ingress[0].ip}')
host=$(kubectl get deployer.core upper -ojsonpath={.status.serviceName})
curl ${ingress} -H "Host: ${host}.default.example.com" -H 'Content-Type: text/plain' -H 'Accept: text/plain' -d riff ; echo
```

PR #177, opened 2019-11-11
Add URL and Address to core Deployer status

scothis:
Further aligns the core deployer with the knative deployer.

Reflects URL and Address for the core Deployer similar to the Knative
Deployer. The URL is used for external access and the address for
in-cluster access.

Refs #142


shashwathi:
@scothis :
 Question about the route URL:  Even if ingress IP is present deployer URL is `http://k8s-square-deployer-9dl5p.default.example.com` is that expected?

PR #181, opened 2019-11-13
Expose Streams as bindable targets

### LINKED ISSUE #180: Expose Streams as bindable services
Streams are like a database in that they can be bound to a workload in order to read/write messages to/from the stream. Following the [Binding spec from CNB](https://github.com/buildpack/spec/blob/master/extensions/bindings.md), we can create a ConfigMap and Secret for each stream that contains the appropriate metadata to be bound into a target workload.

ConfigMap properties:
- `kind`: `'streams.streaming.projectriff.io'`
- `provider`: `stream.name`
- `tags`: empty

Secret properties:
- `gateway`: `stream.status.address.gateway`
- `topic`: `stream.status.address.topic`

The act of binding the ConfigMap/Secret to a workload is out of scope for this work.

scothis:
Create a ConfigMap and Secret to enable bindings based on the CNB
Binding spec. The ConfigMap contains the binding metadata files and the
Secret contains the binding secret files.

ObjectReferences to the ConfigMap and Secret are saved in the Stream's
status.

https://github.com/buildpack/spec/blob/master/extensions/bindings.md

Resolves #180


ericbottard:
See my PR for fixing permissions on secrets.

Also, I would have expected an RFC for this kind of work. Although the PR appears totally fine, I think we need to discuss the aim / how here: what do we do / can we do once we have those cms/secrets? How do we expect to interact with them? How do we "bind"? How does that manifest?  Does that mean a new "binding" CRD? What are the consequences in _e.g._ the CLI? Can we expect to "bind" any type of workload to a stream, including workloads not managed by riff? _etc._

Two other issues I noticed:
- because of the permission issue, the secret bindings were not in effect when I tested. Yet, the stream was marked Ready, I'm not sure I understand how that could be.
- With the permissions fixed, I see secrets created, but can't see their `data` with `kubectl get`. I do get there is something special about `StringData`, but my understanding is that this should end up in `data` like any other secret. I may be missing something, let me know if you witness the same.

markfisher:
+1 for an RFC :-)

scothis:
> I see secrets created, but can't see their data

🤦‍♂ fixed

> I would have expected an RFC for this kind of work

I didn't create an RFC because the RFC processes isn't ratified. I'll work with @trisberg to create a Bindings RFC, but that will have a much broader scope than this PR.

> I think we need to discuss the aim / how here: what do we do / can we do once we have those cms/secrets? How do we expect to interact with them? How do we "bind"? How does that manifest? Does that mean a new "binding" CRD? What are the consequences in e.g. the CLI? Can we expect to "bind" any type of workload to a stream, including workloads not managed by riff?

I intentionally left that out of scope for #180. This is a quick and dirty attempt to make forward progress for workloads to consume streams by building on the CNB Binding spec without getting lost in the broader quagmire of bindings. Nothing changes for Processors, but other types of workloads can manually inject the metadata and secret into their podspec.

If it helps, forget that the word "binding" exists in this PR.

scothis:
@ericbottard @markfisher can you take another look now the RFC-0002 is approved

pkg/controllers/streaming/stream_controller.go
```diff
@@ -82,21 +95,244 @@ func (r *StreamReconciler) reconcile(ctx context.Context, logger logr.Logger, st
...
+	if err != nil {
+		log.Error(err, "unable to reconcile child binding secret Secret", "stream", stream)
+		return ctrl.Result{}, err
+	}
+	stream.Status.Binding.SecretRef = corev1.LocalObjectReference{}
+	if childBindingSecret != nil {
+		stream.Status.Binding.SecretRef.Name = childBindingSecret.Name
+	}
+
+	if childBindingMetadata == nil {
```
markfisher: if both are nil, only metadata will be mentioned in the status message (probably ok)

pkg/controllers/streaming/stream_controller.go
```diff
@@ -82,21 +95,244 @@ func (r *StreamReconciler) reconcile(ctx context.Context, logger logr.Logger, st
...
 	// delegate to the provider via its REST API
-	logger.Info("calling provisioner for Stream", "provisioner", stream.Spec.Provider)
+	log.Info("calling provisioner for Stream", "provisioner", stream.Spec.Provider)
 	address, err := r.StreamProvisionerClient.ProvisionStream(stream)
 	if err != nil {
 		stream.Status.MarkStreamProvisionFailed(err.Error())
 		return ctrl.Result{Requeue: true}, err
 	}
 	stream.Status.MarkStreamProvisioned()
 	stream.Status.Address = *address
```
markfisher: should we just move the ConfigMap/Secret generation here rather than writing the address into the Status, since its duplicating information in the Secret (especially if we think that information does indeed belong in a Secret)?

pkg/controllers/streaming/stream_controller.go
```diff
@@ -82,21 +95,244 @@ func (r *StreamReconciler) reconcile(ctx context.Context, logger logr.Logger, st
...
+			Labels: map[string]string{
+				streamingv1alpha1.StreamLabelKey: stream.Name,
+			},
+			Annotations: make(map[string]string),
+			Name:        fmt.Sprintf("%s-stream-binding-metadata", stream.Name),
+			Namespace:   stream.Namespace,
+		},
+		Data: map[string]string{
+			"kind":     (&streamingv1alpha1.Stream{}).GetGroupVersionKind().GroupKind().String(),
+			"provider": stream.Name,
```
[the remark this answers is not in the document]
ericbottard: No, I think what "provides" those particular values is indeed "the stream A". That it comes from kafka-1 or pulsar-B is irrelevant.
Maybe this warrants clarification in RFC-0001 :)

pkg/controllers/streaming/stream_controller.go
ericbottard: I think yes eventually, but as an interim migration path, this needs to be kept

scothis: I'm not worried about the gateway and topic names as sensitive values. Keeping them for backcompt in the moment is useful, we can decide to remove them before shipping.

We should not add actual credentials to the Stream's status, those should only go into the Secret.

pkg/controllers/streaming/stream_controller.go
scothis: When reconciliation fails, what's is typically reported is the first failure, and not every possible subsequent failure. We could split the condition into separate values for the ConfigMap and Secret, but that seemed like overkill.

pkg/controllers/streaming/stream_controller.go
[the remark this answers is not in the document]
scothis: What provider means is a bit fungible, and I nearly used the stream provider before switching to the stream name. The stream's provider should be irrelevant. Perhaps the provider is the riff streaming runtime.

I asked Ben in the [CNB slack](https://buildpacks.slack.com/archives/CCP60GJLS/p1573662502037400);
```
scothis 11:28 AM
@nebhale can you give an example of 'provider' in the binding spec?
nebhale 11:29 AM
ClearDB for a MySQL kind. (edited) 
scothis 11:56 AM
thanks
```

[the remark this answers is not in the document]
markfisher: okay. I was taking "provider" too literally within _our_ domain, rather than the general CNB spec usage

pkg/controllers/streaming/stream_controller.go
markfisher: I still think we should move the ConfigMap/Secret generation rather than having it cascade from the Address. That way it happens where it would if we didn't already have Address (or if we remove it later). I'm still not sold on having the value in both a Secret and directly in the Status. If it's not sensitive, why not add it to the ConfigMap instead of the Secret? Is this a matter of conforming to the CNB spec's distinction between buildtime and runtime as reflected in the `metadata` vs `secret` sub-directories? If so, I think that should be reconsidered.

scothis: > I still think we should move the ConfigMap/Secret generation rather than having it cascade from the Address.

We can cross that bridge when we get to it. I'd like to build out the capability iteratively.

> If it's not sensitive, why not add it to the ConfigMap instead of the Secret?

Not all values in secrets are sensitive, but all sensitive values should be in secrets.

>  Is this a matter of conforming to the CNB spec's distinction between buildtime and runtime as reflected in the metadata vs secret sub-directories?

The CNB Binding spec doesn't say where an incremental value should go. But there are a couple hints. Here's my current thinking:
1. if the value contains a credential or other sensitive value, it goes under secret
2. if the value is needed at build time it goes under metadata
3. if the value describes what the bound service, it goes under metadata
4. if the value describes how to connect to the bound service, it goes under secret
5. default to the metadata

The current gateway address and topic name fall under item 4, and go into the secret.

markfisher: okay, this sounds fine as long as we have an issue to remove Address from the Status before 0.5

scothis:
Followup work to remove Status Address before 0.5 release in #201

PR #188, opened 2019-11-19
Require a Processor's Streams be ready

### LINKED ISSUE #117: Fail streaming processor reconciliation when stream not ready

scothis:
Sets a Processer's max replicas to zero when any input or output Stream
is not ready. Adds a new status condition StreamsReady which aggregates
all input and output stream ready conditions. The StreamsReady condition
itself is aggregated to the Processor's Ready condition.

Resolves #117
Alternative to #170

### DIFF
--- a/pkg/controllers/streaming/processor_controller.go
+++ b/pkg/controllers/streaming/processor_controller.go
@@ -174,19 +174,19 @@ func (r *ProcessorReconciler) reconcile(ctx context.Context, logger logr.Logger,
 	}
 
 	// Resolve input addresses
-	inputAddresses, _, err := r.resolveStreams(ctx, processorNSName, processor.Spec.Inputs)
+	inputStreams, err := r.resolveStreams(ctx, processorNSName, processor.Spec.Inputs)
 	if err != nil {
 		return ctrl.Result{Requeue: true}, err
 	}
-	processor.Status.InputAddresses = inputAddresses
+	processor.Status.InputAddresses = r.collectStreamAddresses(inputStreams)
 
 	// Resolve output addresses
-	outputAddresses, outputContentTypes, err := r.resolveStreams(ctx, processorNSName, processor.Spec.Outputs)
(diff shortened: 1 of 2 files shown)


pkg/controllers/streaming/processor_controller.go
```diff
@@ -197,6 +197,15 @@ func (r *ProcessorReconciler) reconcile(ctx context.Context, logger logr.Logger,
 	processor.Status.DeploymentName = deployment.Name
 	processor.Status.PropagateDeploymentStatus(&deployment.Status)
 
+	processor.Status.MarkStreamsReady()
+	for _, stream := range append(inputStreams, outputStreams...) {
+		if !stream.Status.IsReady() {
+			ready := stream.Status.GetCondition(stream.Status.GetReadyConditionType())
+			processor.Status.MarkStreamsNotReady(fmt.Sprintf("Stream %s is not ready: %s", stream.Name, ready.Message))
```
ericbottard: In my tests, a nil dereference happens here (ready == nil)

This is with a created Stream whose provisioner address does not respond (because the provider is not there (yet))

scothis: Can we please, please, please get those tests under CI?

scothis: fixed

scothis:
> I think it should set the replica count of the Deployment (not just the ScaledObject) to zero 

The replicas are now set to zero when the Deployment is created. The replica count must be controlled by the scaler.

ericbottard:
Merged as 7041961 alongside my fix.

PR #190, opened 2019-11-19
Preserve additional slashes when splitting gw/topic

### LINKED ISSUE #189: Support slashes in topic names
Pulsar has slashes in topic names.

At least [this piece of code](https://github.com/projectriff/system/blob/80119a4cf2095282c271cb794fff76f3acf6df07/pkg/controllers/streaming/processor_controller.go#L317) will break.

ericbottard:
Fixes #189

### DIFF
--- a/pkg/controllers/streaming/processor_controller.go
+++ b/pkg/controllers/streaming/processor_controller.go
@@ -312,9 +312,9 @@ func triggers(proc *streamingv1alpha1.Processor) []kedav1alpha1.ScaleTriggers {
 	for i, topic := range proc.Status.InputAddresses {
 		result[i].Type = "liiklus"
 		result[i].Metadata = map[string]string{
-			"address": strings.Split(topic, "/")[0],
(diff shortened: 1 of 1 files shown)


pkg/controllers/streaming/processor_controller.go
```diff
@@ -312,9 +312,9 @@ func triggers(proc *streamingv1alpha1.Processor) []kedav1alpha1.ScaleTriggers {
 	for i, topic := range proc.Status.InputAddresses {
 		result[i].Type = "liiklus"
 		result[i].Metadata = map[string]string{
-			"address": strings.Split(topic, "/")[0],
+			"address": strings.SplitN(topic, "/", 2)[0],
```
fbiville: nit: not strictly needed here, right?

ericbottard: No, but: more consistent and I guess, more efficient.

PR #191, opened 2019-11-19
Introduce IngressPolicy on Deployers

### LINKED ISSUE #179: Deployers indicate whether ingress should be created
Add a property to core and knative Deployers indicating whether an Ingress should be created.

Name of property is TBD

scothis:
The IngressPolicy defines whether the workload is exposed outside the
cluster via ingress. There are two policies:

- `ClusterLocal`: no ingress
- `External`: creates ingress

`ClusterLocal` is the default policy, which requires that authors
explicitly opt-in to exposing their workload outside of the cluster. Set
the policy to `External` to access the workload outside the cluster.

Note: the cluster local policy is not sufficient to protect workload
from external network access if the cluster is also not adequately
protected. It merely prevents the workload from being registered via
kubernetes ingress.

Resolves #179


scothis:
cc @joshrider

Issue #198, opened 2019-11-22
Consume stream bindings in Processor


ericbottard:

ericbottard:
@scothis Assigning to you since you mentioned it in Slack, so that we don't overlap. Feel free to un-assign.

ericbottard:
Feels like the functionality of reconciling the `bindings:` part of a workload to create volume mounts should be a library (maybe not external to this repo at first), and the `Processor` reconciler would leverage it (in the same way that an hypothetical `Source` reconciler would use it), right?

ericbottard:
Implemented in d62ae0d3060f0528e090ec4607ab40ec78dde2f3

PR #202, opened 2019-11-22
Processor stream bindings

scothis:
Consume stream bindings on the Processor's processor container. The processor image still needs to be updated in order to consume the bindings.

Once the processor image is updated, fields marked by this PR as deprecated can be removed.

Refs #198, [RFC 0002](https://github.com/projectriff/riff/blob/master/rfc/rfc-0002-bindings.md)

### DIFF
--- a/pkg/controllers/streaming/processor_controller.go
+++ b/pkg/controllers/streaming/processor_controller.go
@@ -178,15 +178,15 @@ func (r *ProcessorReconciler) reconcile(ctx context.Context, logger logr.Logger,
 	if err != nil {
 		return ctrl.Result{Requeue: true}, err
 	}
-	processor.Status.InputAddresses = r.collectStreamAddresses(inputStreams)
+	processor.Status.DeprecatedInputAddresses = r.collectStreamAddresses(inputStreams)
 
 	// Resolve output addresses
 	outputStreams, err := r.resolveStreams(ctx, processorNSName, processor.Spec.Outputs)
 	if err != nil {
 		return ctrl.Result{Requeue: true}, err
 	}
-	processor.Status.OutputAddresses = r.collectStreamAddresses(outputStreams)
-	processor.Status.OutputContentTypes = r.collectStreamContentTypes(outputStreams)
+	processor.Status.DeprecatedOutputAddresses = r.collectStreamAddresses(outputStreams)
+	processor.Status.DeprecatedOutputContentTypes = r.collectStreamContentTypes(outputStreams)
 
 	// Reconcile deployment for processor
 	deployment, err := r.reconcileProcessorDeployment(ctx, logger, processor, &cm)
@@ -326,8 +326,8 @@ func (r *ProcessorReconciler) constructScaledObjectForProcessor(processor *strea
 }
 
 func triggers(proc *streamingv1alpha1.Processor) []kedav1alpha1.ScaleTriggers {
-	result := make([]kedav1alpha1.ScaleTriggers, len(proc.Status.InputAddresses))
-	for i, topic := range proc.Status.InputAddresses {
+	result := make([]kedav1alpha1.ScaleTriggers, len(proc.Status.DeprecatedInputAddresses))
+	for i, topic := range proc.Status.DeprecatedInputAddresses {
 		result[i].Type = "liiklus"
 		result[i].Metadata = map[string]string{
 			"address": strings.SplitN(topic, "/", 2)[0],
@@ -521,7 +521,7 @@ func (r *ProcessorReconciler) collectStreamContentTypes(streams []streamingv1alp
 }
 
 func (r *ProcessorReconciler) computeEnvironmentVariables(processor *streamingv1alpha1.Processor) ([]v1.EnvVar, error) {
-	contentTypesJson, err := json.Marshal(processor.Status.OutputContentTypes)
+	contentTypesJson, err := json.Marshal(processor.Status.DeprecatedOutputContentTypes)
 	if err != nil {
 		return nil, err
 	}
@@ -530,11 +530,11 @@ func (r *ProcessorReconciler) computeEnvironmentVariables(processor *streamingv1
 	return []v1.EnvVar{
 		{
 			Name:  "INPUTS",
-			Value: strings.Join(processor.Status.InputAddresses, ","),
+			Value: strings.Join(processor.Status.DeprecatedInputAddresses, ","),
 		},
 		{
 			Name:  "OUTPUTS",
-			Value: strings.Join(processor.Status.OutputAddresses, ","),
+			Value: strings.Join(processor.Status.DeprecatedOutputAddresses, ","),
 		},
 		{
 			Name:  "INPUT_NAMES",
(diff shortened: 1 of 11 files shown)


pkg/controllers/streaming/processor_controller.go
```diff
@@ -422,6 +426,53 @@ func (r *ProcessorReconciler) constructDeploymentForProcessor(processor *streami
 		return nil, err
 	}
 
+	volumes := []corev1.Volume{}
+	volumeMounts := []corev1.VolumeMount{}
+	for _, stream := range streams {
+		if stream.Status.Binding.MetadataRef.Name != "" {
+			metadataVolumeName := fmt.Sprintf("processor-stream-%s-metadata", stream.Name)
```
ericbottard: I don't think the name here needs to contain processor, does it? When/if this is extracted as a library function, the fact that the binding is from a processor, or a source or whatever doesn't matter. Once stream X is bound, it's the same data available however it is consumed.

pkg/controllers/streaming/processor_controller.go
```diff
@@ -422,6 +426,53 @@ func (r *ProcessorReconciler) constructDeploymentForProcessor(processor *streami
...
+							LocalObjectReference: corev1.LocalObjectReference{
+								Name: stream.Status.Binding.MetadataRef.Name,
+							},
+						},
+					},
+				},
+			)
+			volumeMounts = append(volumeMounts,
+				corev1.VolumeMount{
+					Name:      metadataVolumeName,
```
ericbottard: Same remark as above, with even the additional comment/discussion below:

shouldn't the "role" played by the stream appear here somehow, ie "first input" (input0), "second input" (input1), etc?
It seems otherwise we need an extra info anyway, I guess conveyed by an ENV VAR.

pkg/controllers/streaming/processor_controller.go
scothis: The name of the volume is only used to coordinate between the volume definition and the mount. I used the prefix to reduce the risk of colliding with an existing volume for the function container in the pod. The name is not visible inside the container at runtime.

pkg/controllers/streaming/processor_controller.go
scothis: See above comment. The name of the volume is not exposed inside the container.

> shouldn't the "role" played by the stream appear here somehow, ie "first input" (input0), "second input" (input1), etc?

I didn't want to introduce an encoding scheme into the name of the binding.

> It seems otherwise we need an extra info anyway, I guess conveyed by an ENV VAR.

Bingo. That's why I didn't mark the INPUT_NAMES and OUTPUT_NAMES envvar for removal.

ericbottard: > The name of the volume is not exposed inside the container

I was thinking more of part of the mount path.

> Bingo. That's why I didn't mark the INPUT_NAMES and OUTPUT_NAMES envvar for removal.

That's a misunderstanding of what INPUT/OUTPUT_NAMES are. They're aliases, logical names that appear in the target function code (eg keys in a map, names of fields in a struct, etc) whose _default_ value come from stream names, but they're NOT stream names.

So it seems we'll always need extra info that the binding spec doesn't address. The question is where do we (and are _we_ the ones to decide, or is it actually mandated in the spec, really? this is why I wanted to discuss this subject further in the RFC, but discussing it over actual running code is fine too) draw the line.

pkg/controllers/streaming/processor_controller.go
ericbottard: I think you missed my point: this piece of code should eventually be extracted as re-usable. The name could be totally opaque for what it's worth, and I think we agree on that. So all I'm saying is that inserting "processor" in it here confuses the reader on that fact.

scothis: We're building this iteratively. When a library is abstracted out we can deal with library concerns. Until then, I'm not concerned with library concerns.

pkg/controllers/streaming/processor_controller.go
scothis: OK, the binding name should be using the alias rather than the stream's name. I'll fix that.

> So it seems we'll always need extra info that the binding spec doesn't address.

That's by design. The binding is not responsible for carying metadata about how the binding should be consumed, it only contains metadata about the provisioned resource that can be accessed. Metadata about how the resource should be consumed must be passed outside of the binding. The spec says:

> The binding name is the only aspect of the binding that is under the control of the consumer. All other elements of the binding are defined by the binding producer. Additional metadata may be injected into the workload about the binding out of band, but is not part of this RFC.

pkg/controllers/streaming/processor_controller.go
```diff
@@ -428,7 +428,9 @@ func (r *ProcessorReconciler) constructDeploymentForProcessor(processor *streami
 
 	volumes := []corev1.Volume{}
 	volumeMounts := []corev1.VolumeMount{}
-	for _, stream := range streams {
+	streams := append(inputStreams, outputStreams...)
```
ericbottard: This can modify the underlying array of `inputStreams` as a side effect. While this is harmless here, I think it may become problematic if `inputStreams` is later re-used and could lead to subtle bug.
Same goes for the line below.

pkg/controllers/streaming/processor_controller.go
```diff
@@ -422,6 +426,55 @@ func (r *ProcessorReconciler) constructDeploymentForProcessor(processor *streami
 		return nil, err
 	}
 
+	volumes := []corev1.Volume{}
+	volumeMounts := []corev1.VolumeMount{}
+	streams := append(inputStreams, outputStreams...)
+	streamBindings := append(processor.Spec.Inputs, processor.Spec.Outputs...)
+	for i, stream := range streams {
+		if stream.Status.Binding.MetadataRef.Name != "" {
+			metadataVolumeName := fmt.Sprintf("processor-stream-%s-metadata", stream.Name)
```
ericbottard: Same line, different argument:
Because it should be possible to do things like `riff streaming processor create ... --output out --output someAlias:out`, the name here should either
- **also** be based on the alias
- stay as the stream name, but be de-duped.

scothis: ok. let's change the parameter to be the alias instead of the stream's name, and add validation that ensures alias uniqueness.

ericbottard: > let's change the parameter to be the alias instead of the stream's name, and add validation that ensures alias uniqueness.

Actually, I think the second option is the only reasonable one, because it should be ok to have the same alias twice, as long as one is on the input side and the other on the output side.

ericbottard: Which causes a problem in the current design because the alias is used in the path.

Is it ok per the spec to use different root paths, including per-type (streams) but even per-role (ie `inputsBindingsRootPath vs outputsBindingsRootPath`) ?

pkg/controllers/streaming/processor_controller.go
scothis: done

pkg/controllers/streaming/processor_controller.go
scothis: > it should be ok to have the same alias twice, as long as one is on the input side and the other on the output side.

I don't think it's unreasonable to have the alias be unique for both inputs and outputs. We can relax this down the road if we deconflict the file path.

> Is it ok per the spec to use different root paths

There is only one directory to look at as far as the CNB Bindings spec is concerned. We could do something different, but then we're deviating from CNB Bindings.

ericbottard:
Merged as d62ae0d

PR #204, opened 2019-11-23
WIP: test streaming

sbawaska:


.github/workflows/run-tests.sh
```diff
@@ -43,5 +43,56 @@ if [ $RUNTIME = "core" ] || [ $RUNTIME = "knative" ]; then
...
+  kubectl create clusterrolebinding dev-util-stream --clusterrole=riff-streaming-readonly-role --serviceaccount=default:default
+
+  for test in java java-boot; do
+    name=system-streaming-cluster-repeater-${test}
+    image=$(fats_image_repo ${name})
+
+    riff function create $name --image $image --namespace $NAMESPACE --tail \
+          --git-repo https://github.com/${FATS_REPO}.git --git-revision ${FATS_REFSPEC} --sub-path functions/repeater/${test} &
+
+    riff streaming stream create letters --namespace $NAMESPACE --provider franz-kafka-provisioner --content-type 'text/plain'
```
scothis: The value of `--provider` is generated, it needs to be looked up from the cluster at runtime

.github/workflows/run-tests.sh
```diff
@@ -43,5 +43,56 @@ if [ $RUNTIME = "core" ] || [ $RUNTIME = "knative" ]; then
...
+      echo "actual_data: $actual_data"
+      if [ "$actual_data" == "[foo, foo]" ]; then
+        echo "Check succedded!"
+        return 0
+      fi
+      sleep 1
+      cnt=$((cnt+1))
+    done
+
+  done
```
scothis: delete the dev-util pod, function, streams, provider and Kafka once done with them

.github/workflows/run-tests.sh
```diff
@@ -43,5 +43,56 @@ if [ $RUNTIME = "core" ] || [ $RUNTIME = "knative" ]; then
...
+  helm repo add incubator http://storage.googleapis.com/kubernetes-charts-incubator
+  helm install --name my-kafka incubator/kafka --set replicas=1,zookeeper.replicaCount=1,zookeeper.env.ZK_HEAP_SIZE=128m --namespace $NAMESPACE --wait
+
+  riff streaming kafka-provider create franz --bootstrap-servers my-kafka:9092 --namespace $NAMESPACE
+
+  cat <<EOF | kubectl apply -f -
+apiVersion: v1
+kind: Pod
+metadata:
+  name: dev-utils
```
scothis:
```suggestion
  name: dev-utils
  namespace: ${NAMESPACE}
```

.github/workflows/run-tests.sh
```diff
@@ -43,5 +43,56 @@ if [ $RUNTIME = "core" ] || [ $RUNTIME = "knative" ]; then
...
+spec:
```
scothis: avoid granting extra permission to the default service account. We wouldn't want uses to do that, and we don't want to give our workloads extra privileges they shouldn't have.

Accepting this suggestion will also necessitate, creating the service account and updating the role binding.

```suggestion
spec:
  serviceAccountName: dev-utils
```

.github/workflows/run-tests.sh
```diff
@@ -43,5 +43,56 @@ if [ $RUNTIME = "core" ] || [ $RUNTIME = "knative" ]; then
...
+  containers:
+  - name: dev-utils
+    image: projectriff/dev-utils
+EOF
+  kubectl create clusterrolebinding dev-util-stream --clusterrole=riff-streaming-readonly-role --serviceaccount=default:default
```
scothis: right-size privilege:

- scope to only the target namespace
- 

```suggestion
  kubectl create rolebinding --namespace $NAMESPACE dev-utils --clusterrole=view --serviceaccount=${NAMESPACE}:dev-utils
```

.github/workflows/run-tests.sh
```diff
@@ -43,5 +43,56 @@ if [ $RUNTIME = "core" ] || [ $RUNTIME = "knative" ]; then
...
+    riff streaming stream create letters --namespace $NAMESPACE --provider franz-kafka-provisioner --content-type 'text/plain'
+    riff streaming stream create numbers --namespace $NAMESPACE --provider franz-kafka-provisioner --content-type 'application/json'
+    riff streaming stream create result --namespace $NAMESPACE --provider franz-kafka-provisioner --content-type 'text/plain'
+
+    riff streaming processor create $name --function-ref $name --namespace $NAMESPACE --input letters --input numbers --output result --tail
+
+    kubectl exec dev-utils -- publish letters --payload foo
+    kubectl exec dev-utils -- publish numbers --payload 2 --content-type "application/json"
+    ${FATS_DIR}/install.sh jq
+    kubectl exec dev-utils -- publish letters --payload foo > result.txt &
```
scothis: should this be watching the result stream instead?

.github/workflows/run-tests.sh
```diff
@@ -43,5 +43,56 @@ if [ $RUNTIME = "core" ] || [ $RUNTIME = "knative" ]; then
...
+    while [ $cnt -lt 60 ]; do
+      cat result.txt.out
+      actual_data=`cat result.out | jq -r .payload`
+      echo "actual_data: $actual_data"
+      if [ "$actual_data" == "[foo, foo]" ]; then
+        echo "Check succedded!"
+        return 0
```
scothis: returning here will prevent other tests from running

.github/workflows/run-tests.sh
```diff
@@ -43,5 +43,56 @@ if [ $RUNTIME = "core" ] || [ $RUNTIME = "knative" ]; then
...
+    riff streaming stream create result --namespace $NAMESPACE --provider franz-kafka-provisioner --content-type 'text/plain'
+
+    riff streaming processor create $name --function-ref $name --namespace $NAMESPACE --input letters --input numbers --output result --tail
+
+    kubectl exec dev-utils -- publish letters --payload foo
+    kubectl exec dev-utils -- publish numbers --payload 2 --content-type "application/json"
+    ${FATS_DIR}/install.sh jq
+    kubectl exec dev-utils -- publish letters --payload foo > result.txt &
+    while [ $cnt -lt 60 ]; do
+      cat result.txt.out
```
scothis: what's the difference between `result.txt`, `result.out` and `result.txt.out`?

.github/workflows/run-tests.sh
```diff
@@ -43,5 +43,56 @@ if [ $RUNTIME = "core" ] || [ $RUNTIME = "knative" ]; then
...
+    riff function create $name --image $image --namespace $NAMESPACE --tail \
+          --git-repo https://github.com/${FATS_REPO}.git --git-revision ${FATS_REFSPEC} --sub-path functions/repeater/${test} &
+
+    riff streaming stream create letters --namespace $NAMESPACE --provider franz-kafka-provisioner --content-type 'text/plain'
+    riff streaming stream create numbers --namespace $NAMESPACE --provider franz-kafka-provisioner --content-type 'application/json'
+    riff streaming stream create result --namespace $NAMESPACE --provider franz-kafka-provisioner --content-type 'text/plain'
+
+    riff streaming processor create $name --function-ref $name --namespace $NAMESPACE --input letters --input numbers --output result --tail
+
+    kubectl exec dev-utils -- publish letters --payload foo
```
scothis: does this need a `--content-type` flag as well?

.github/workflows/run-tests.sh
```diff
@@ -43,5 +43,56 @@ if [ $RUNTIME = "core" ] || [ $RUNTIME = "knative" ]; then
...
+  # TODO move setup to fats
+  ${FATS_DIR}/install.sh helm
+  source ${FATS_DIR}/macros/helm-init.sh
+
+  helm repo add incubator http://storage.googleapis.com/kubernetes-charts-incubator
+  helm install --name my-kafka incubator/kafka --set replicas=1,zookeeper.replicaCount=1,zookeeper.env.ZK_HEAP_SIZE=128m --namespace $NAMESPACE --wait
+
+  riff streaming kafka-provider create franz --bootstrap-servers my-kafka:9092 --namespace $NAMESPACE
+
+  cat <<EOF | kubectl apply -f -
```
scothis: creating (and deleting) the dev-utils pod, along with the service account and role binding would make a good FATS macro.

.github/workflows/run-tests.sh
```diff
@@ -43,5 +43,56 @@ if [ $RUNTIME = "core" ] || [ $RUNTIME = "knative" ]; then
...
+apiVersion: v1
+kind: Pod
+metadata:
+  name: dev-utils
+spec:
+  containers:
+  - name: dev-utils
+    image: projectriff/dev-utils
```
scothis: use a tag other than latest for stability

sbawaska:
closing in favor of https://github.com/projectriff/system/pull/210

Issue #151, opened 2019-10-28
Streaming processor scales to zero even when there is constant activity   


jldec:
This was observed in a riff setup with core, knative, and streaming runtimes installed using `--devel` build via `helm install` on oct 28, 2019.

A random number generator posts a number to an input stream via the streaming gateway every second. The stream is consumed by a simple echo function deployed as a streaming processor. 

It appears that keda is scaling this processor down to 0 every 30s or so, even though there is constant activity on the input stream

echo.js
```js
module.exports = x => {
  console.log('echo', x);
  return x;
}
```

```
# deploy gateway (yaml uses riff-system namespace)
kubectl apply -f https://storage.googleapis.com/projectriff/riff-http-gateway/riff-http-gateway-0.5.0-snapshot.yaml

# create in and out streams
riff streaming stream create in  --provider franz-kafka-provisioner --content-type application/json 
riff streaming stream create out --provider franz-kafka-provisioner --content-type application/json 

# build echo function
riff function create echo \
  --local-path ~/riff/echo \
  --artifact echo.js \
  --tail

# deploy echo function as a stream processor 
riff streaming processor create echo-out \
  --function-ref echo \
  --input in \
  --output out

# create random number generator from image
riff container create random --image jldec/random:v0.0.2 --tail

# deploy random number generator as core-random using core runtime
riff core deployer create core-random --container-ref random --tail

# (in new terminal) port forward to core-random on port 8081
kubectl port-forward service/`kubectl get service -l core.projectriff.io/deployer=core-random -o jsonpath='{.items[0].metadata.name}'` 8081:80

# configure random to send a number to the in stream every second via the http-gateway
curl http://localhost:8081/ -w '\n' \
-H 'Content-Type: application/json' \
-d '{"url":"http://riff-streaming-http-gateway.riff-system/default/in"}'

# monitor the input stream using the liiklus client - it should show a new number on the stream every second. The name of the Kafka provider is franz 

# (in new terminal) port forward to kafka-provider-liiklus on port 6565
kubectl port-forward service/`kubectl get service -l streaminprojectriff.io/kafka-provider-liiklus=franz -o jsonpath='{.items[0].metadata.name}'` 6565:6565

# (in a new terminal) watch in the input stream 
# (this assumes that the liiklus client has been built at ~/riff/liiklus-client)
java -jar ~/riff/liiklus-client/target/liiklus-client-1.0.0-SNAPSHOT.jar --consumer localhost:6565 default_in
```

[...]

PR #210, opened 2019-11-27
Test streaming functions

scothis:


scothis:
@sbawaska @ericbottard I've stripped this PR down to the core necessary to run the streaming runtime in CI, and added a bunch of extra logging. The tests are failing because no messages are emitted from the processor. The processor Pod is being shutdown before it starts.

- the KEDA cooldown period is ignored. Jurgen reported a different manifestation of the same issue in #151. Are we doing something wrong, or is there a bug in KEDA? We set the cool down period to 30 seconds.
- are we ACKing the message in the processor sidecar before remitting the message to the function? There should still be lag on the topic until the function is able to handle the message (success or fail). Just because a message is read, doesn't mean that work is complete.

ericbottard:
> the KEDA cooldown period is ignored. Jurgen reported a different manifestation of the same issue in #151. Are we doing something wrong, or is there a bug in KEDA? We set the cool down period to 30 seconds.

We're not doing anything wrong (on the riff side). The keda liiklus scaler (and any scaler in general) doesn't take into account that a scaler should be considered "Active" when there is *progress* (in addition to obvious *lag*). Sadly, to be able to track progress (ie the fact that current offset is further now than it was one tick ago), the scaler API would need to maintain state, which is really not how KEDA is architectured at the moment.

> are we ACKing the message in the processor sidecar before remitting the message to the function? There should still be lag on the topic until the function is able to handle the message (success or fail). Just because a message is read, doesn't mean that work is complete.

We are indeed ACKing the message one-by-one in the processor (for now, because ultimately we'll need to give more control to the function). I'm not sure we can push it down some more (in particular, we can't decide to ACK based on *output*, because there is no one-to-one mapping between function input messages and function outputs)

ericbottard:
That being said, for a function that sees a single message at time t0, it should stay up till t0+30s, even today.

ericbottard:
May be nice to later change the tested function to do things only possible via streaming (eg emit N results) but LGTM in the current form

Issue #193, opened 2019-11-20
Make domain configurable for core deployer


trisberg:
Currently we hard-code `example.com` as the domain, we should make that configurable

andrew-su:
Currently configurable from ConfigMap, but we may want to allow configuration during install rather than post install.

PR #213, opened 2019-11-28
Add support to modify the ingress' default domain from a config map

andrew-su:
This PR adds a new configmap `riff-core-default-domain` which can be used to configure the default domain specified.

Example config map:
```
apiVersion: v1
data:
  defaultDomain: myexample.com
kind: ConfigMap
metadata:
  name: riff-core-default-domain
  namespace: default
```

The default is still `example.com`

Related Issues:
https://github.com/projectriff/system/issues/193
https://github.com/projectriff/system/issues/142

TODO: Make this configurable via helm install.


config/riff-build.yaml
```diff
@@ -8,205 +8,201 @@ metadata:
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
-  annotations:
```
scothis: this should still be present. You may need to update your local copy of controller-gen. If you delete you copy and run make, it will install the correct version.

pkg/controllers/core/deployer_controller.go
```diff
@@ -613,8 +636,37 @@ func (r *DeployerReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		Owns(&networkingv1beta1.Ingress{}).
 		// watch for build mutations to update dependent deployers
+		Watches(&source.Kind{Type: &corev1.ConfigMap{}}, r.enqueueTrackedConfigMapResource()).
 		Watches(&source.Kind{Type: &buildv1alpha1.Application{}}, enqueueTrackedResources(&buildv1alpha1.Application{})).
 		Watches(&source.Kind{Type: &buildv1alpha1.Container{}}, enqueueTrackedResources(&buildv1alpha1.Container{})).
 		Watches(&source.Kind{Type: &buildv1alpha1.Function{}}, enqueueTrackedResources(&buildv1alpha1.Function{})).
 		Complete(r)
 }
+
+func (r *DeployerReconciler) constructHostForDeployer(deployer *corev1alpha1.Deployer, cm corev1.ConfigMap) string {
```
scothis: move to be closer to where this method is used. Just below `constructLabelsForDeployer` would be a good spot.

pkg/controllers/core/deployer_controller.go
```diff
@@ -613,8 +636,37 @@ func (r *DeployerReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	domain := defaultDomain
+	if d, ok := cm.Data[defaultDomainKey]; ok {
```
scothis: handle case when valve is empty

```suggestion
	if d := cm.Data[defaultDomainKey]; d != "" {
```

pkg/controllers/core/deployer_controller.go
```diff
@@ -613,8 +636,37 @@ func (r *DeployerReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		domain = d
+	}
+
+	return fmt.Sprintf("%s.%s", deployer.Name, domain)
+}
+
+func (r *DeployerReconciler) enqueueTrackedConfigMapResource() handler.EventHandler {
```
scothis: let's enhance the current  enqueueTrackedResource function to support configmaps rather than redefine it

pkg/controllers/core/deployer_controller.go
```diff
@@ -613,8 +636,37 @@ func (r *DeployerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.Service{}).
 		Owns(&networkingv1beta1.Ingress{}).
 		// watch for build mutations to update dependent deployers
+		Watches(&source.Kind{Type: &corev1.ConfigMap{}}, r.enqueueTrackedConfigMapResource()).
```
scothis:
```suggestion
		Watches(&source.Kind{Type: &corev1.ConfigMap{}}, enqueueTrackedResources(&corev1.ConfigMap{})).
```

pkg/controllers/core/deployer_controller.go
```diff
@@ -357,7 +379,7 @@ func (r *DeployerReconciler) constructPodSpecForDeployer(deployer *corev1alpha1.
 	return podSpec
 }
 
-func (r *DeployerReconciler) reconcileIngress(ctx context.Context, log logr.Logger, deployer *corev1alpha1.Deployer, serviceName string) (*networkingv1beta1.Ingress, error) {
+func (r *DeployerReconciler) reconcileIngress(ctx context.Context, log logr.Logger, deployer *corev1alpha1.Deployer, serviceName string, cm corev1.ConfigMap) (*networkingv1beta1.Ingress, error) {
```
scothis:
```suggestion
func (r *DeployerReconciler) reconcileIngress(ctx context.Context, log logr.Logger, deployer *corev1alpha1.Deployer, serviceName string, cm *corev1.ConfigMap) (*networkingv1beta1.Ingress, error) {
```

pkg/controllers/core/deployer_controller.go
```diff
@@ -103,11 +103,33 @@ func (r *DeployerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	return result, err
 }
 
+func (r *DeployerReconciler) getAndTrackConfigMap(ctx context.Context, log logr.Logger, deployer *corev1alpha1.Deployer) (corev1.ConfigMap, error) {
```
scothis:
```suggestion
func (r *DeployerReconciler) getAndTrackConfigMap(ctx context.Context, log logr.Logger, deployer *corev1alpha1.Deployer) (*corev1.ConfigMap, error) {
```

pkg/controllers/core/deployer_controller.go
```diff
@@ -103,11 +103,33 @@ func (r *DeployerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	cmKey := types.NamespacedName{Namespace: deployer.Namespace, Name: coreDeployerConfigMapName}
+
+	// track config map for updated domain name
+	r.Tracker.Track(
+		tracker.NewKey(schema.GroupVersionKind{Version: "v1", Kind: "ConfigMap"}, cmKey),
+		types.NamespacedName{Namespace: deployer.Namespace, Name: deployer.Name},
+	)
+	if err := r.Get(ctx, cmKey, &cm); err != nil {
+		log.Error(err, "unable to lookup images configMap")
+		return cm, err
```
scothis:
```suggestion
		return nil, err
```

pkg/controllers/core/deployer_controller.go
```diff
@@ -103,11 +103,33 @@ func (r *DeployerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	}
+
+	return cm, nil
```
scothis:
```suggestion
	return &cm, nil
```

pkg/controllers/core/config.go
```diff
@@ -0,0 +1,25 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package core
+
+const (
+	kustomizePrefix = "riff-core" // kustomize adds this prefix to all our resource names
+
+	coreDeployerConfigMapName = kustomizePrefix + "-default-domain"
```
scothis: Let's use a single ConfigMap for all config rather than a ConfigMap for each key.

```suggestion
	configMapName = kustomizePrefix
```

cmd/managers/core/main.go
```diff
@@ -39,6 +39,7 @@ var (
 	scheme     = runtime.NewScheme()
 	setupLog   = ctrl.Log.WithName("setup")
 	syncPeriod = 10 * time.Hour
+	namespace  = os.Getenv("SYSTEM_NAMESPACE")
```
scothis: rm

I appreciate the intent, and making the namespace dynamic is something we should presume eventually, however, it is also a much bigger effort that we should think about separately as it has downstream impact.

config/core/manager/manager.yaml
```diff
@@ -18,6 +18,11 @@ spec:
       containers:
       - args:
         - --enable-leader-election
+        env:
```
scothis: rm

cmd/managers/core/main.go
andrew-su: Is the currently expected namespace to be always `riff-system`? This was to follow the example set forth by streaming + build to detect the config's namespace via the env defined in the yaml.

For now I've just hard coded the system namespace as "riff-system" in the config.

pkg/controllers/core/config.go
andrew-su: I've updated to use a more generic name. The config map generated by kustomize is always prefixed with `riff-core-`. The config is now named `riff-core-settings`.

pkg/controllers/core/deployer_controller.go
andrew-su: After thinking about this more, the config map resource is different from the other ones. We only want to watch for changes to the config map with a specific name in the system namespace. Whereas the generic `enqueueTrackedResource` tracks all items of the kind passed in.

pkg/controllers/core/config.go
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package core
+
+const (
+	kustomizePrefix = "riff-core" // kustomize adds this prefix to all our resource names
+
+	systemNamespace = "riff-system"
+
+	configMapName    = kustomizePrefix + "-settings"
```
scothis:
```suggestion
	settingsConfigMapName = kustomizePrefix + "-settings"
```

pkg/controllers/core/deployer_controller.go
scothis: That's the role of the Tracker. The `ToRequests` method will be called for every event for the watched kind. The tracker won't be tracking any resources for most ConfigMaps, so there is nothing to queue. For the target ConfigMap, every deployer will be tracking it, so every deployer will be added to the reconciliation queue.

andrew-su: Updated!

config/core/config/kustomization.yaml
```diff
@@ -0,0 +1,2 @@
+resources:
+  - ingress-domain.yaml
```
scothis: Also need to rename the target file.

```suggestion
  - settings.yaml
```

pkg/controllers/core/deployer_controller.go
```diff
@@ -103,11 +101,39 @@ func (r *DeployerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	}
+
+	cmKey := types.NamespacedName{Namespace: systemNamespace, Name: settingsConfigMapName}
+
+	r.Tracker.Track(
+		tracker.NewKey(versionKinds[0], cmKey),
+		types.NamespacedName{Namespace: deployer.Namespace, Name: deployer.Name},
+	)
+	if err := r.Get(ctx, cmKey, cm); err != nil {
+		log.Error(err, "unable to lookup images configMap")
```
scothis: the message is stale now that this is a generic method

pkg/controllers/core/deployer_controller.go
andrew-su: If a lookup happens for something not being tracked, it will log that the item is not being tracked. The logs get a lot of these messages.

pkg/controllers/core/deployer_controller.go
```diff
@@ -103,11 +101,39 @@ func (r *DeployerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	return result, err
 }
 
+func (r *DeployerReconciler) getAndTrackConfigMap(ctx context.Context, log logr.Logger, deployer *corev1alpha1.Deployer) (*corev1.ConfigMap, error) {
+	cm := &corev1.ConfigMap{}
+
+	versionKinds, _, err := r.Scheme.ObjectKinds(cm)
```
scothis: we only need to lookup the kind for a generic Object, since we know this is a ConfigMap the kind will always be `ConfigMap` and the apiVersion will always be `v1`.

pkg/controllers/core/deployer_controller.go
```diff
@@ -103,11 +101,39 @@ func (r *DeployerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	return result, err
 }
 
+func (r *DeployerReconciler) getAndTrackConfigMap(ctx context.Context, log logr.Logger, deployer *corev1alpha1.Deployer) (*corev1.ConfigMap, error) {
```
scothis: No need to couple this method to the deployer. Any object can be tracked.

```suggestion
func (r *DeployerReconciler) getAndTrackConfigMap(ctx context.Context, log logr.Logger, obj runtime.Object) (*corev1.ConfigMap, error) {
```

It may also be worth pulling this method out of this reconciler, because it could be generally useful.

pkg/controllers/core/deployer_controller.go
```diff
@@ -103,11 +101,39 @@ func (r *DeployerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+
+	return cm, nil
+}
+
 func (r *DeployerReconciler) reconcile(ctx context.Context, log logr.Logger, deployer *corev1alpha1.Deployer) (ctrl.Result, error) {
 	if deployer.GetDeletionTimestamp() != nil {
 		return ctrl.Result{}, nil
 	}
 
+	cm, err := r.getAndTrackConfigMap(ctx, log, deployer)
```
scothis: rename `cm` to something meaningful.

pkg/controllers/core/deployer_controller.go
```diff
@@ -580,17 +607,32 @@ func (r *DeployerReconciler) constructLabelsForDeployer(deployer *corev1alpha1.D
 	return labels
 }
 
+func (r *DeployerReconciler) constructHostForDeployer(deployer *corev1alpha1.Deployer, cm *corev1.ConfigMap) string {
+	domain := defaultDomain
+	if d := cm.Data[defaultDomainKey]; d != "" {
+		domain = d
+	}
+
+	return fmt.Sprintf("%s.%s", deployer.Name, domain)
```
scothis:
```suggestion
	return fmt.Sprintf("%s.%s.%s", deployer.Name, deployer.Namespace, domain)
```

pkg/controllers/core/deployer_controller.go
```diff
@@ -580,17 +607,32 @@ func (r *DeployerReconciler) constructLabelsForDeployer(deployer *corev1alpha1.D
...
-					t.GetGroupVersionKind(),
-					types.NamespacedName{Namespace: a.Meta.GetNamespace(), Name: a.Meta.GetName()},
-				)
-				for _, item := range r.Tracker.Lookup(key) {
-					requests = append(requests, reconcile.Request{NamespacedName: item})
+				var requests []reconcile.Request
+
+				versionKinds, _, err := r.Scheme.ObjectKinds(t)
+				if err != nil {
+					r.Log.Error(err, "Unable to detect version kind")
```
scothis: let's panic here, it should never happen and is indicative of a programming error.

pkg/controllers/core/deployer_controller.go
```diff
@@ -580,17 +607,32 @@ func (r *DeployerReconciler) constructLabelsForDeployer(deployer *corev1alpha1.D
...
+				} else {
```
scothis: remove else, and outdent

pkg/controllers/core/deployer_controller.go
andrew-su: I've reverted this back to the version before with the explicit version and kind.

pkg/controllers/core/deployer_controller.go
andrew-su: Perhaps as a separate change? Right now, we're tracking a specific configmap `riff-core-settings`.  As the refactor should probably pull out the `getAndTrackConfigMap` function from the other controllers as well, I want to hold out on that and keep this change limited to just the core deployer.

scothis: I'd still like to decouple this method from the Deployer type since the runtime.Object interface has everything we need.

Perhaps the method should also be renamed to `getAndTrackSettings` since the method doesn't get any arbitrary config map.

pkg/controllers/core/deployer_controller.go
scothis: the message still needs to be updated

pkg/controllers/core/deployer_controller.go
```diff
@@ -103,11 +102,33 @@ func (r *DeployerReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+
+	return cm, nil
+}
+
 func (r *DeployerReconciler) reconcile(ctx context.Context, log logr.Logger, deployer *corev1alpha1.Deployer) (ctrl.Result, error) {
 	if deployer.GetDeletionTimestamp() != nil {
 		return ctrl.Result{}, nil
 	}
 
+	configMap, err := r.getAndTrackConfigMap(ctx, log, deployer)
```
scothis: `configMap` is still too generic. What's the semantics role of the config map? The updated name should also be updated within the reconcileIngress and constructIngressForDeployer methods

Issue #215, opened 2019-11-29
Improve diagnostics for deployer service name collisions


glyn:
@scothis identified some improvements: https://github.com/projectriff/system/pull/214#pullrequestreview-324752547

glyn:
For example, create two deployers with the same service name:
```
$ kubectl create namespace issue215
$ riff container create test-container --image glyn/kubernetes-bootcamp:v1 -n issue215
$ riff core deployer create test-deployer --container-ref test-container -n issue215 --service-name test-service
$ riff core deployer create test-deployer2 --container-ref test-container -n issue215 --service-name test-service
```

The conflicting deployer doesn't make the conflict obvious:
```
$ kubectl describe deployer test-deployer2 -n issue215
Name:         test-deployer2
Namespace:    issue215
Labels:       <none>
Annotations:  <none>
API Version:  core.projectriff.io/v1alpha1
Kind:         Deployer
Metadata:
  Creation Timestamp:  2019-11-29T16:09:13Z
  Generation:          1
  Resource Version:    2526367
  Self Link:           /apis/core.projectriff.io/v1alpha1/namespaces/issue215/deployers/test-deployer2
  UID:                 95e09bcd-12c2-11ea-b90f-42010a840047
Spec:
  Build:
    Container Ref:  test-container
  Ingress Policy:   External
  Service Name:     test-service
  Template:
    Containers:
      Name:  handler
      Ports:
        Container Port:  8080
        Name:            http
        Protocol:        TCP
      Resources:
Status:
  Conditions:
    Last Transition Time:  2019-11-29T16:09:18Z
    Status:                True
    Type:                  DeploymentReady
    Last Transition Time:  2019-11-29T16:09:14Z
    Message:               ReplicaSet "test-deployer2-deployer-4prss-845db65dc6" is progressing.
    Reason:                ReplicaSetUpdated
    Status:                Unknown
    Type:                  Ready
    Last Transition Time:  2019-11-29T16:09:13Z
    Status:                Unknown
    Type:                  ServiceReady
  Deployment Name:         test-deployer2-deployer-4prss
  Latest Image:            glyn/kubernetes-bootcamp:v1
Events:                    <none>
```

PR #217, opened 2019-12-02
Handle deployer service name collisions - WIP

glyn:
Fixes https://github.com/projectriff/system/issues/215

### DIFF
--- a/pkg/apis/core/v1alpha1/deployer_lifecycle.go
+++ b/pkg/apis/core/v1alpha1/deployer_lifecycle.go
@@ -17,6 +17,8 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"fmt"
+
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	networkingv1beta1 "k8s.io/api/networking/v1beta1"
@@ -25,10 +27,10 @@ import (
 )
 
 const (
-	DeployerConditionReady                              = apis.ConditionReady
-	DeployerConditionDeploymentReady apis.ConditionType = "DeploymentReady"
-	DeployerConditionServiceReady    apis.ConditionType = "ServiceReady"
-	DeployerConditionIngressReady    apis.ConditionType = "IngressReady"
+	DeployerConditionReady                                   = apis.ConditionReady
+	DeployerConditionDeploymentReady      apis.ConditionType = "DeploymentReady"
+	DeployerConditionServiceReady         apis.ConditionType = "ServiceReady"
+	DeployerConditionIngressReady         apis.ConditionType = "IngressReady"
 )
 
 var deployerCondSet = apis.NewLivingConditionSet(
@@ -87,6 +89,10 @@ func (ds *DeployerStatus) PropagateDeploymentStatus(cds *appsv1.DeploymentStatus
 func (ds *DeployerStatus) PropagateServiceStatus(ss *corev1.ServiceStatus) {
 	// services don't have meaningful status
 	deployerCondSet.Manage(ds).MarkTrue(DeployerConditionServiceReady)
+}
+
+func (ds *DeployerStatus) MarkServiceNameInUse(collisionKind string, collisionName string) {
+	deployerCondSet.Manage((ds)).MarkFalse(DeployerConditionServiceReady, "serviceNameCollision", fmt.Sprintf("kind %s, name %s", collisionKind, collisionName))
 }
 
 func (ds *DeployerStatus) MarkIngressNotRequired() {
--- a/pkg/controllers/core/deployer_controller.go
+++ b/pkg/controllers/core/deployer_controller.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"fmt"
 
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
 	"github.com/go-logr/logr"
 	"github.com/google/go-cmp/cmp"
 	appsv1 "k8s.io/api/apps/v1"
@@ -132,6 +134,7 @@ func (r *DeployerReconciler) reconcile(ctx context.Context, log logr.Logger, dep
 	childService, err := r.reconcileChildService(ctx, log, deployer)
 	if err != nil {
 		log.Error(err, "unable to reconcile child Service", "deployer", deployer)
+		// FIXME: capture error in ServiceStatus false message
 		return ctrl.Result{}, err
 	}
 	deployer.Status.ServiceName = childService.Name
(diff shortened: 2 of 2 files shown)


pkg/apis/core/v1alpha1/deployer_lifecycle.go
```diff
@@ -89,6 +92,10 @@ func (ds *DeployerStatus) PropagateServiceStatus(ss *corev1.ServiceStatus) {
 	deployerCondSet.Manage(ds).MarkTrue(DeployerConditionServiceReady)
 }
 
+func (ds *DeployerStatus) ServiceNameCollision(collisionKind string, collisionName string) {
```
scothis: methods that manipulate a condition typically take the form `Mark{kind}NotOwned`

See https://github.com/projectriff/system/blob/v0.4.x/pkg/apis/core/v1alpha1/deployer_lifecycle.go#L56-L59

pkg/controllers/core/deployer_controller.go
```diff
@@ -557,6 +558,38 @@ func (r *DeployerReconciler) constructServiceForDeployer(deployer *corev1alpha1.
 	// If a service name was provided, use it, otherwise generate one based on the deployer name.
 	if deployer.Spec.ServiceName != "" {
 		service.ObjectMeta.Name = deployer.Spec.ServiceName
+
+		// Detect service name collision.
+		collidingService, err := r.getService(deployer.Namespace, deployer.Spec.ServiceName)
```
scothis: rather than proactive detection, let the create fail. Even if this tests says the resource doesn't exist, it could be created before the controller calls create (and fails anyway)

pkg/controllers/core/deployer_controller.go
```diff
@@ -557,6 +558,38 @@ func (r *DeployerReconciler) constructServiceForDeployer(deployer *corev1alpha1.
...
+		if err != nil {
+			return nil, err
+		}
+		if collidingService != nil {
+			// Diagnose service name collisions. If the colliding service was created by a deployer, include that in the
```
scothis: it doesn't matter why the resource collided. The only thing that matters is the resource already exists and the deployer being reconciled doesn't own it.

pkg/controllers/core/deployer_controller.go
```diff
@@ -557,6 +558,38 @@ func (r *DeployerReconciler) constructServiceForDeployer(deployer *corev1alpha1.
...
+					break
+				}
+			}
+			err := fmt.Errorf("service name provided on deployer collided with existing service %v", collidingService)
+			r.Log.Error(err, "collision", "deployer", deployer, "collidingService", collidingService,
+				"collisionKind", collisionKind, "collisionName", collisionName)
+			deployer.Status.ServiceNameCollision(collisionKind, collisionName)
+
+			// Track the colliding service for collision removal.
+			r.Tracker.Track(
```
scothis: tracking the collided resource is an anti-pattern in this case. If that resource is part of a production system, imagine an upgrade needs to delete and then recreate the same resource (some fields are immutable). If this controller is watching for the resource to be deleted, we now have a race for who can create the resource first.

pkg/controllers/core/deployer_controller.go
```diff
@@ -616,5 +675,18 @@ func (r *DeployerReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Watches(&source.Kind{Type: &buildv1alpha1.Application{}}, enqueueTrackedResources(&buildv1alpha1.Application{})).
 		Watches(&source.Kind{Type: &buildv1alpha1.Container{}}, enqueueTrackedResources(&buildv1alpha1.Container{})).
 		Watches(&source.Kind{Type: &buildv1alpha1.Function{}}, enqueueTrackedResources(&buildv1alpha1.Function{})).
+		// watch for service mutations to update colliding deployers
+		Watches(&source.Kind{Type: &corev1.Service{}}, enqueueResourcesCollidingWithServices()).
```
scothis: `Owns` is a more specific form of `Watches`, since Service is already "owned" this isn't necessary.

glyn:
> I don't think we should continue down this path. See [projectriff/riff#1369](https://github.com/projectriff/riff/pull/1369) for my thoughts on why. tl;dr we should drop ServiceName as a user configurable property.

Ok, let's discuss at today's sync.

> 
> I left comments on the specifics of this PR as if we should continue.

Thanks for that as I'm still learning kubernetes. I'll respond in the same spirit.

pkg/apis/core/v1alpha1/deployer_lifecycle.go
glyn: So would you prefer something like `MarkServiceNameInUse`?

pkg/controllers/core/deployer_controller.go
glyn: The idea is to provide better diagnostics for the user to take action on. Currently, apart from the service not being ready, there's no indication that the reason is because the service name was already in use. But maybe we can let the create fail and then add suitable diagnostics to the `ServiceReady` condition in this case. wdyt?

pkg/controllers/core/deployer_controller.go
glyn: Ok, so capturing the fact of a collision should give the user sufficient information to fix the problem.

pkg/controllers/core/deployer_controller.go
glyn: Good point about non-determinism. But if we don't track the collided resource, how will we know when to re-reconcile if the collided resource if modified or deleted to remove the collision?

pkg/controllers/core/deployer_controller.go
glyn: Interesting. I thought that `Owns` would only track services which are owned by deployers. The reason for using `Watches` too was to track colliding services which are not owned by deployers. I know you don't want to track "unowned" services, but if we did decide to do that, would `Owns` achive it?

pkg/apis/core/v1alpha1/deployer_lifecycle.go
```diff
@@ -25,15 +26,17 @@ import (
...
 
 const (
-	DeployerConditionReady                              = apis.ConditionReady
-	DeployerConditionDeploymentReady apis.ConditionType = "DeploymentReady"
-	DeployerConditionServiceReady    apis.ConditionType = "ServiceReady"
-	DeployerConditionIngressReady    apis.ConditionType = "IngressReady"
+	DeployerConditionReady                                   = apis.ConditionReady
+	DeployerConditionDeploymentReady      apis.ConditionType = "DeploymentReady"
+	DeployerConditionServiceReady         apis.ConditionType = "ServiceReady"
+	DeployerConditionServiceNameAvailable apis.ConditionType = "ServiceNameAvailable"
```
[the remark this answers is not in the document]
glyn: Yes, it seems reasonable to overload `ServiceReady` since the service can't be created if its name is already in use.

pkg/controllers/core/deployer_controller.go
glyn: I guess my proposal isn't bullet-proof because a service creation can fail for some reason but then someone else could immediately create a service with the desired name and we can't assume that the failure was due to the name being in use.

Issue #223, opened 2019-12-03
Make the service name used by a core deployer equal to the deployer name


glyn:
See https://github.com/projectriff/system/issues/200#issuecomment-561230330 for context.

Issue #200, opened 2019-11-22
Make the service name used by a core deployer configurable


trisberg:
Currently the deployers generate a service name to keep them unique. This makes it harder to discover services since you can't bind using just the DNS name, you would have to look it up based on labels. We should provide a way to define the actual name used and the end-user would be responsible for avoiding clashes. If user doesn't specify the name, then we should generate one as we do today.

jldec:
Good idea, i like it. Was there a reason not to use the deployer name as the default service name?

jldec:
Example with proposed CLI flag
`riff core deployer create inventory --expose-service-as mysvc`

If flag is not provided, the current (auto-generated) name is used.
Defaulting to match the deployer name can result in conflicts so avoiding that for now.

glyn:
@scothis is questioning this approach - re-opening for further discussion.

glyn:
The guidance that is being developed in [RFC 0006](https://github.com/projectriff/riff/pull/1369) suggests that we generate a static name (maybe even without suffix) as user-specified names for auto-generated resources create a number of edge cases.

So let's revert the earlier implementation of this issue and follow up with another issue.

PR #224, opened 2019-12-03
Make the service name used by a core deployer equal to the deployer name

glyn:
Diagnose service name collisions by capturing any service creation error in the relevant condition.

See https://github.com/projectriff/system/pull/141 for where dynamically generated service names were first introduced.

Fixes https://github.com/projectriff/system/issues/223

Reverts "Make the service name used by the deployer configurable".

This reverts commit 09c47935c6ab01cde9db164feb38561af5056fca.

Closes https://github.com/projectriff/system/issues/200


glyn:
Here's an example of the Status of a deployer whose service name already existed:
```
Status:
  Conditions:
    Last Transition Time:  2019-12-03T16:21:14Z
    Status:                True
    Type:                  DeploymentReady
    Last Transition Time:  2019-12-03T16:20:59Z
    Message:               services "test-deployer" already exists
    Reason:                reconcileError
    Status:                False
    Type:                  Ready
    Last Transition Time:  2019-12-03T16:20:59Z
    Message:               services "test-deployer" already exists
    Reason:                reconcileError
    Status:                False
    Type:                  ServiceReady
  Deployment Name:         test-deployer-deployer-7qbj4
  Latest Image:            glyn/kubernetes-bootcamp:v1
Events:                    <none>
```

pkg/controllers/core/deployer_controller.go
```diff
@@ -132,6 +132,7 @@ func (r *DeployerReconciler) reconcile(ctx context.Context, log logr.Logger, dep
 	childService, err := r.reconcileChildService(ctx, log, deployer)
 	if err != nil {
 		log.Error(err, "unable to reconcile child Service", "deployer", deployer)
+		deployer.Status.MarkServiceReconcileError(err)
```
scothis: we only want to update the condition if the service resource fails to be created because the name already exists.

By always setting the condition for any error, transient errors will causes the ready condition of the deployer to turn false. The riff cli with the `--tail` flag will watch the resource until the ready condition is no longer unknown. This change as proposed will cause the CLI to report failure much more frequently.

pkg/apis/core/v1alpha1/deployer_lifecycle.go
```diff
@@ -89,6 +89,10 @@ func (ds *DeployerStatus) PropagateServiceStatus(ss *corev1.ServiceStatus) {
 	deployerCondSet.Manage(ds).MarkTrue(DeployerConditionServiceReady)
 }
 
+func (ds *DeployerStatus) MarkServiceReconcileError(err error) {
```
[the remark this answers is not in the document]
glyn: I'll adopt this. I overlooked that method.

pkg/controllers/core/deployer_controller.go
glyn: Makes sense. Also, that will make the condition reason `createError` instead of the rather bland `reconcileError`.

pkg/apis/core/v1alpha1/deployer_lifecycle.go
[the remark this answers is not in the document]
glyn: Actually, I see that riff 0.4 was checking services ahead of reconciliation, which is a design we've moved away from. Rather than trying to introspect the service creation error, which is fragile, I think it's sufficient to capture the service creation error in the service ready condition.

pkg/controllers/core/deployer_controller.go
glyn: Actually made the `ServiceReady` condition reason `serviceCreationFailed` so it makes more sense when percolated up to the `Ready` condition reason.

PR #228, opened 2019-12-04
Add min and max scale to Knative Deployer

joshrider:
Adds Scale to the DeployerSpec. When set, knative's
autoscaler will work within the defined Minimum and
Maximum.

Addresses #1368


pivotal-issuemaster:
@joshrider Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=projectriff/system&pullRequestId=228)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=projectriff/system&pullRequestId=228) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@joshrider Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=projectriff/system&pullRequestId=228)!

joshrider:
How do I placate the robot?

pkg/apis/knative/v1alpha1/deployer_types.go
```diff
@@ -64,6 +66,11 @@ const (
 	IngressPolicyExternal     IngressPolicy = "External"
 )
 
+type Scale struct {
+	Minimum *int64 `json:"minimum,omitempty"`
+	Maximum *int64 `json:"maximum,omitempty"`
```
scothis: int64 may be a wee bit optimistic about our ability to scale

pkg/apis/knative/v1alpha1/deployer_types.go
```diff
@@ -64,6 +66,11 @@ const (
 	IngressPolicyExternal     IngressPolicy = "External"
 )
 
+type Scale struct {
+	Minimum *int64 `json:"minimum,omitempty"`
```
scothis: I'd tend to lean towards `min` and `max`

pkg/apis/knative/v1alpha1/deployer_validation.go
```diff
@@ -88,6 +88,19 @@ func (s DeployerSpec) Validate() validation.FieldErrors {
 		errs = errs.Also(validation.ErrInvalidValue(s.IngressPolicy, "ingressPolicy"))
 	}
 
+	if s.Scale != nil {
+		if s.Scale.Minimum != nil && *s.Scale.Minimum < int64(0) {
```
scothis: move this into a Validate method on the Scale struct. That will help the tests since the scale values can be tested independent of the rest of the spec.

pkg/apis/knative/v1alpha1/deployer_types.go
```diff
@@ -45,6 +45,8 @@ type DeployerSpec struct {
 	// deployer.
 	Build *Build `json:"build,omitempty"`
 
+	Scale *Scale `json:"scale,omitempty"`
```
scothis: Since nil has no semantic meaning, we can simplify access by not using a pointer

```suggestion
	Scale Scale `json:"scale,omitempty"`
```

config/build/crd/bases/build.projectriff.io_applications.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.2.4
+    controller-gen.kubebuilder.io/version: (devel)
```
scothis: controller-gen v0.2.4 is required. If you delete the version on your path, running make will install the correct version.

pkg/apis/knative/v1alpha1/deployer_types.go
joshrider: still being optimistic with an `*int32`, but matching knative's type for Desired/Actual Scale

pkg/apis/knative/v1alpha1/deployer_validation.go
joshrider: Moved these and also broke the Scale validation tests out to their own test. LMK if I've gone too far.

pkg/apis/knative/v1alpha1/deployer_validation_test.go
```diff
@@ -125,3 +125,65 @@ func TestValidateDeployerSpec(t *testing.T) {
 		})
 	}
 }
+
+func TestValidateScale(t *testing.T) {
```
scothis: let's add one scaling test case that produces an error back to TestValidateDeployerSpec just to confirm that the spec is calling scale.Validate() and reporting the errors under the expected field path.

pkg/controllers/knative/deployer_controller.go
```diff
@@ -414,6 +418,25 @@ func (r *DeployerReconciler) constructLabelsForDeployer(deployer *knativev1alpha
...
+	// make a little extra space just in case
+	annotations := make(map[string]string, len(deployer.ObjectMeta.Annotations)+2)
+
+	// pass through existing annotations
+	for k, v := range deployer.ObjectMeta.Annotations {
+		annotations[k] = v
+	}
+
+	if deployer.Spec.Scale.Min != nil {
+		annotations["autoscaling.knative.dev/minScale"] = strconv.Itoa(int(*deployer.Spec.Scale.Min))
```
scothis: consider something like this to avoid type downcasting, or switch the type in the resource from int32 to int

```suggestion
		annotations["autoscaling.knative.dev/minScale"] = fmt.Sprintf("%d", *deployer.Spec.Scale.Min)
```

pkg/apis/knative/v1alpha1/deployer_validation_test.go
joshrider: Added a test to the DeployerSpec validation 👍

pkg/controllers/knative/deployer_controller.go
joshrider: Change made! Is the idea that casting is a heavy activity and something to be avoided where possible?

PR #230, opened 2019-12-05
adding a readyness probe to streaming processor

sbawaska:


pkg/controllers/streaming/processor_controller.go
```diff
@@ -532,6 +533,18 @@ func (r *ProcessorReconciler) constructDeploymentForProcessor(processor *streami
...
+		ReadinessProbe: &v1.Probe{
+			Handler: v1.Handler{
+				TCPSocket: &v1.TCPSocketAction{
+					Port: intstr.IntOrString{
+						IntVal: 8081,
+					},
+				},
+			},
+			InitialDelaySeconds: 2,
+			TimeoutSeconds:      300,
```
scothis: 300 is a very long timeout. I'd lean towards using the default probe values unless we have a compelling reason to be different. If so, let's document that reason.

sbawaska: Ok, using the default for TimeoutSeconds.

scothis: what about InitialDelaySeconds and PeriodSeconds?

sbawaska: I could not find the default value for `InitialDelaySeconds`, but since we are waiting for a JVM to start, I felt 2 seconds was appropriate. Also, `PeriodSeconds` has a default of 10 seconds, which I felt was too big.

scothis: The readiness probe doesn't need to be super precise since there isn't a service sending traffic to the processor. It's really more of an informational signal to those watching the resource.

pkg/controllers/streaming/processor_controller.go
```diff
@@ -532,6 +533,15 @@ func (r *ProcessorReconciler) constructDeploymentForProcessor(processor *streami
 		ImagePullPolicy: v1.PullIfNotPresent,
 		Env:             environmentVariables,
 		VolumeMounts:    volumeMounts,
+		ReadinessProbe: &v1.Probe{
+			Handler: v1.Handler{
+				TCPSocket: &v1.TCPSocketAction{
+					Port: intstr.IntOrString{
+						IntVal: 8081,
```
scothis: it's a bit odd to have a container's probe reference a port claimed by a different container in the pod. Each container should have it's own probes

pkg/controllers/streaming/processor_controller.go
```diff
@@ -532,6 +533,15 @@ func (r *ProcessorReconciler) constructDeploymentForProcessor(processor *streami
 		ImagePullPolicy: v1.PullIfNotPresent,
 		Env:             environmentVariables,
 		VolumeMounts:    volumeMounts,
+		ReadinessProbe: &v1.Probe{
+			Handler: v1.Handler{
+				TCPSocket: &v1.TCPSocketAction{
+					Port: intstr.IntOrString{
```
scothis: favor the helper methods, like `intstr.FromInt()`

sbawaska: TIL!

pkg/controllers/streaming/processor_controller.go
sbawaska: duh! fixed.

pkg/controllers/streaming/processor_controller.go
```diff
@@ -526,6 +527,16 @@ func (r *ProcessorReconciler) constructDeploymentForProcessor(processor *streami
...
 		},
 	}
+	tcpProbe := v1.Probe{
+		Handler: v1.Handler{
+			TCPSocket: &v1.TCPSocketAction{
+				Port: intstr.FromInt(8081),
+			},
+		},
+	}
+	podSpec.Containers[0].ReadinessProbe = &tcpProbe
```
scothis: if the user defines a probe, we shouldn't overwrite it

scothis:
stale

PR #231, opened 2019-12-05
[Hold-RFC0005] Introduce OffsetReset on input binding.

### LINKED ISSUE #208: Expose OffsetReset on the Processor CRD (and reconciler)

ericbottard:
Fixes #208


ericbottard:
Rebased on top of latest master

pkg/apis/streaming/v1alpha1/processor_types.go
```diff
@@ -75,6 +75,23 @@ type StreamBinding struct {
...
+type InputStreamBinding struct {
+	// Stream name, from this namespace, to be bound to the processor
+	Stream string `json:"stream"`
+
+	// Alias exposes the stream under another name within the processor
+	// +optional
+	Alias string `json:"alias,omitempty"`
+
+	// Where to start consuming this stream the first time a processor runs.
+	StartOffset string `json:"startOffset"`
```
fbiville: missing `// +optional`

fbiville: It works with that fix

ericbottard: Should work without it as there is a defaulter. Moreover, this is only a spec and code snippet was here to give a "feeling" of the change. Better to discuss this in the `system` PR if at all

fbiville: We are in the system PR ^^
Re-tested this morning without the `// +optional` and it works now.

fbiville:
Merged as b4a03f4e0d8c2c2b51896aa13f2e196428b79d57 🎅

PR #232, opened 2019-12-05
Processor controller integration test - WIP

glyn:
The purpose of this PR is two-fold:
1. agree an approach to testing our controllers/reconcilers
2. add tests for a specific controller/reconciler.

### DIFF
--- a/config/thirdparty/keda/keda.k8s.io_scaledobjects_crd.yaml
+++ b/config/thirdparty/keda/keda.k8s.io_scaledobjects_crd.yaml
@@ -1,0 +1,4529 @@
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  name: scaledobjects.keda.k8s.io
+spec:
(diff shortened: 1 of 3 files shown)


glyn:
> I'm still a bit skeptical of using a real api server. The value of unit tests is being able to inject different types of faults and making sure the reconciler behaves in expected ways. The happy path is largely covered by acceptance tests (or will be for streaming).

Agreed. I think it was worth investigating the current approach since it was recommended by kubebuilder, but it does seem quite tricky to get right. Also, I'd love to avoid flakes such as `etcd` being slow to start, etc.

> I'd like to also spike out tests that follow the Knative reconciler tests that setup given state and can assert every create/update/delete query.

Yes, I'm looking into those. There's a steeper learning curve, but it looks like the resultant tests are more likely to be maintainable. I'm not sure how easy it is to inject errors.

glyn:
Closing. See https://github.com/projectriff/riff/pull/1372 for rationale.

PR #233, opened 2019-12-06
Adopt Knative approach to reconciler testing - WIP

glyn:
The purpose of this PR is two-fold:
1. agree an approach to testing our controllers/reconcilers
2. add tests for a specific controller/reconciler.


glyn:
See https://github.com/projectriff/riff/pull/1372 for rationale.

PR #236, opened 2019-12-07
Use TypedLocalObjectReferences in status

scothis:
Following the advice of [RFC 0007](https://github.com/projectriff/riff/blob/master/rfc/rfc-0007-resource-references.md) to use structured references, this PR switches away from bare resources names in the status for our CRDs.

This is a breaking change for tools expecting to consume the name of a child resource from the parent's status. For consumers, the change follows the pattern `.status.*Name` -> `.status.*Ref.name`.

TypedLocalObjectReferences are used, which include the apiGroup, kind and name. Which is less than ObjectReferences contains, but is enough to unambiguously resolve the resource.

Due to an incompatibility with the controller-gen generated openapi schema and the marshaled resource, I've had to vendor and hack the TypedLocalObjectReferences. Our vendor type marshals identically to corev1.TypedLocalObjectReferences.


pkg/refs/refs.go
```diff
@@ -0,0 +1,97 @@
...
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+)
+
+func NewTypedLocalObjectReference(name string, gk schema.GroupKind) *TypedLocalObjectReference {
+	if name == "" || gk.Empty() {
+		return nil
```
ericbottard: Is it really sound to return nil in all those edge cases rather than say, panic? Seems like we never check for nil at call sites

ericbottard:
LGTM after we discuss the checks for nil issue

pkg/refs/refs.go
scothis: I agree this is being overly defensive.

ericbottard:
Merged as 1a0ef53

PR #237, opened 2019-12-09
Properly Watch() the processor image ConfigMap

### LINKED ISSUE #209: Processor controller Tracks() CM but doesn't Watch() it

ericbottard:
Fixes #209

Without this fix, changing the value in the ConfigMap did not trigger
a change in the deployed processors.

### DIFF
--- a/pkg/controllers/streaming/processor_controller.go
+++ b/pkg/controllers/streaming/processor_controller.go
@@ -76,7 +76,9 @@ type ProcessorReconciler struct {
 // +kubebuilder:rbac:groups=keda.k8s.io,resources=scaledobjects,verbs=get;list;watch;create;update;patch;delete
 // Watches
 // +kubebuilder:rbac:groups=streaming.projectriff.io,resources=streams,verbs=get;watch
-// +kubebuilder:rbac:groups=build.projectriff.io,resources=containers;functions,verbs=get;watch
+// +kubebuilder:rbac:groups=build.projectriff.io,resources=containers,verbs=get;watch
+// +kubebuilder:rbac:groups=build.projectriff.io,resources=functions,verbs=get;watch
+// +kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch
 
 func (r *ProcessorReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	ctx := context.Background()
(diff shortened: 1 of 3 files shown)


pkg/controllers/streaming/processor_controller.go
```diff
@@ -682,6 +684,21 @@ func (r *ProcessorReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			}),
 		}
 	}
+	enqueueNamedTrackedConfigMap := &handler.EnqueueRequestsFromMapFunc{
```
scothis: Let's combine this with enqueueTrackedResource similar to what @joshrider came up with:
https://github.com/projectriff/system/blob/28aac56da45362d4852822be5f93c2825a539533/pkg/controllers/core/deployer_controller.go#L614-L635

joshrider: credit: @andrew-su

ericbottard: Not sure I get what you mean. Are you asking for me to *also* replace `enqueueTrackedResources` that already exists above with the one referenced (and keep `enqueueNamedTrackedConfigMap`) or to come up with a unique function that covers both cases (maybe with an optional name for example)?

scothis: we can replace both with the linked method.

> maybe with an optional name for example

I don't follow

ericbottard: The one I added has `&& a.Meta.GetName() == processorImages` in the condition.

scothis: ah yes. That's a performance optimization, but is still correct without that check.

I'd like to eventually pull this method into a separate utility so we don't need to repeat it everywhere. If you don't make that change in this PR, we can pick it up then.

ericbottard: OK, changed to use that one function.

Notice how I've moved the 
```go
		versionKinds, _, err := r.Scheme.ObjectKinds(t)
		if err != nil {
			panic(err)
		}
```
bit out of the callback though.

pkg/controllers/streaming/processor_controller.go
```diff
@@ -667,12 +669,16 @@ func (*ProcessorReconciler) collectAliases(bindings []streamingv1alpha1.StreamBi
...
+		if err != nil {
+			panic(err)
+		}
 		return &handler.EnqueueRequestsFromMapFunc{
 			ToRequests: handler.ToRequestsFunc(func(a handler.MapObject) []reconcile.Request {
-				requests := []reconcile.Request{}
+				var requests []reconcile.Request
 				key := tracker.NewKey(
-					t.GetGroupVersionKind(),
+					versionKinds[0],
```
ericbottard: Maybe we should also panic above if len(versionKinds) != 1

scothis: > Maybe we should also panic above if len(versionKinds) != 1

There can be multiple valid GroupVersions for a single type. Ingress is in `networking.k8s.io/v1beta1 ` and `extensions/v1beta1`.

ericbottard:
Merged as 1377163

PR #238, opened 2019-12-09
Controller resolves container image digest

### LINKED ISSUE #227: Implement Container image digesting
The Container resources passes the image value directly to the status's latest image. We should:
- resolve the image to contain a digest using ggcr
- poll for new images

It doesn't seem possible to change the resync interval for a single type of resource, however, we should be able to create an external heartbeat that adds containers to the work queue.

ekcasey:
Fixes #227

* LatestImage is populated with a resolved digest reference
* TargetImage is now fully explicit
* Introduces a new keychain the resolves registry auth from annotated secrets


ekcasey:
Question: it seems like the annotation `kpack` (and with this PR the container controller) need on docker secrets are only applied by the CLI. Should we apply these in the CredentialReconciler if they are missing? Without these annotations components of `riff-build` won't work against private repositories.

pkg/authn/secrets_keychain_test.go
```diff
@@ -0,0 +1,270 @@
...
+			},
+			expected: gauthn.Anonymous,
+		},
+		{
+			name: "gcr secret",
+			secrets: []v1.Secret{
+				{
+					ObjectMeta: v12.ObjectMeta{
+						Annotations: map[string]string{
+							"build.pivotal.io/docker": "https://gcr.io",
```
scothis:
```suggestion
							DockerSecretAnnotation: "https://gcr.io",
```

pkg/controllers/build/container_controller.go
```diff
@@ -129,6 +154,70 @@ func (r *ContainerReconciler) resolveTargetImage(ctx context.Context, log logr.L
...
+		log.Error(err, "failed to get image digest", "image", ref.String())
+		return "", err
+	}
+
+	return fmt.Sprintf("%s@%s", ref.Context().Name(), digest), nil
+}
+
+func (r *ContainerReconciler) constructKeychain(ctx context.Context, log logr.Logger, container *buildv1alpha1.Container) (gauthn.Keychain, error) {
+	var serviceAccount v1.ServiceAccount
+	if err := r.Get(ctx, types.NamespacedName{Namespace: container.Namespace, Name: riffBuildServiceAccount}, &serviceAccount); err != nil {
```
scothis: We need to define RBAC policies and an Informer to be watch ServiceAccounts and Secrets. This currently happens to work because other controllers are already consumes these resources, but we should as well in case this reconciler is ever run in isolation, or the other reconciler is removed.

For the RBAC policy, we need to add at line ~50:
```
// +kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch
// +kubebuilder:rbac:groups=core,resources=serviceaccounts,verbs=get;list;watch
```

For the Informer, we need to add in SetupWithManager, between the `For` and `Complete`:
```
		Watches(&source.Kind{Type: &corev1.Secret{}}, handler.Funcs{}).
		Watches(&source.Kind{Type: &corev1.ServiceAccount{}}, handler.Funcs{}).
```

pkg/authn/secrets_keychain.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+type DockerSecretsKeychain struct {
+	regAuths map[string]*gauthn.Basic
+}
+
+const DockerSecretAnnotation = "build.pivotal.io/docker"
+
+var aliases = map[string][]string{
+	"gcr.io": {"us.gcr.io", "eu.gcr.io", "asia.gcr.io"},
```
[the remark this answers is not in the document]
ekcasey: I removed the aliases. We should consider if we want to make the keychain resolve for generic subdomains in the future.

pkg/controllers/build/container_controller.go
```diff
@@ -129,8 +158,74 @@ func (r *ContainerReconciler) resolveTargetImage(ctx context.Context, log logr.L
...
+		return "", err
+	}
+
+	digest, err := img.Digest()
+	if err != nil {
+		log.Error(err, "failed to get image digest", "image", ref.String())
+		return "", err
+	}
+
+	return fmt.Sprintf("%s@%s", ref.Context().Name(), digest), nil
```
glyn: Will this preserve any tag from the input reference? Since `ref.Context()` returns a _repository_ reference, I guess that would strip out any tag.

glyn:
One option to reduce the amount of code in this PR is to depend on github.com/pivotal/image-relocation rather than github.com/google/go-containerregistry. The former repository provides a simple way to get the digest of an image plus some rich types for manipulating image references without resorting to string manipulation.

pkg/controllers/build/container_controller.go
ekcasey: @glyn It does strip out the tag. `ref.Name()` includes the tag for a tag reference but `ref.Context().Name()` shouldn't include the tag or digest.

glyn: Ok, so the question is: is that what is required? It's architecturally ok to have a tag plus a digest (although some tooling can't cope with it) and sometimes preserving the digest is helpful for human readability.

ekcasey: What gets set on `LatestImage` with the given implementation should be a digest reference w/o a tag which seems to me to be safer choice (given that as you mention some tooling doesn't like having both a tag and a digest).

ekcasey:
@scothis I changed this PR to essentially poll the registry for images updates once a minute. Without this change, users may be stuck using an outdated digest after pushing updates to the provided tag. I imagine we might like to make the polling interval configurable eventually. Please take a look and let me know what you think.

ekcasey:
@glyn I looked into using github.com/pivotal/image-relocation. Because it doesn't appear that we can inject a keychain I don't think it will work for us in this case (https://github.com/pivotal/image-relocation/blob/dc3df03f826ed15b6400976ac7f91f6a70e8569a/pkg/registry/ggcr/remote.go#L34).

glyn:
> @glyn I looked into using github.com/pivotal/image-relocation. Because it doesn't appear that we can inject a keychain I don't think it will work for us in this case (https://github.com/pivotal/image-relocation/blob/dc3df03f826ed15b6400976ac7f91f6a70e8569a/pkg/registry/ggcr/remote.go#L34).

Ok, thanks for investigating. Maybe we could put that feature into image-relocation and then replace this code in due course. I don't mind how you prefer to play this.

pkg/controllers/build/container_controller.go
```diff
@@ -129,8 +164,74 @@ func (r *ContainerReconciler) resolveTargetImage(ctx context.Context, log logr.L
...
+}
+
+func (r *ContainerReconciler) constructKeychain(ctx context.Context, log logr.Logger, container *buildv1alpha1.Container) (gauthn.Keychain, error) {
+	var serviceAccount corev1.ServiceAccount
+	if err := r.Get(ctx, types.NamespacedName{Namespace: container.Namespace, Name: riffBuildServiceAccount}, &serviceAccount); err != nil {
+		if apierrs.IsNotFound(err) {
+			log.Info("service account not found", "service-account", riffBuildServiceAccount)
+			return gauthn.DefaultKeychain, nil
```
scothis: We dynamically create the `riff-build` service account when we see a credential, function or application. Since containers now need the service account as well, we should also setup the service account when a container is created.

That logic, exists here:
https://github.com/projectriff/system/blob/1377163d1d9832dad87c73e33a9284af1ec3542d/pkg/controllers/build/credentials_controller.go#L252-L254

Here we should fail and requeue the reconciliation because the service account should become available shortly.

pkg/controllers/build/container_controller.go
```diff
@@ -95,24 +107,47 @@ func (r *ContainerReconciler) reconcile(ctx context.Context, log logr.Logger, co
...
 
-	// TODO resolve to a digest
-	container.Status.LatestImage = container.Status.TargetImage
+	container.Status.LatestImage = latestImage
 
 	container.Status.ObservedGeneration = container.Generation
 
-	return ctrl.Result{}, nil
+	return ctrl.Result{
+		Requeue:      true,
```
scothis: Requeue is implicitly true when RequeueAfter is set

https://godoc.org/sigs.k8s.io/controller-runtime/pkg/reconcile#Result

PR #243, opened 2019-12-13
Check controller-gen version

fbiville:


fbiville:
Giving up for now. Tried on Mac OS (where it works) and Ubuntu (where it fails as in CI). No idea what's wrong with my changes...

PR #247, opened 2019-12-17
Initial Application controller tests

scothis:
- basic tests for the Application controller
  - create kpack image
  - update status from a ready kpack iamge
  - update status from a non-ready kpack iamge
- copy created resources when logging, before calling hook or wrapped client
- deterministically generate resource names on create

### DIFF
--- a/pkg/controllers/build/application_controller_test.go
+++ b/pkg/controllers/build/application_controller_test.go
@@ -1,0 +1,389 @@
+/*
+Copyright 2019 the original author or authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


pkg/controllers/build/application_controller_test.go
```diff
@@ -0,0 +1,389 @@
...
+						Git: &buildv1alpha1.Git{
+							URL:      testGitUrl,
+							Revision: testGitRevision,
+						},
+					},
+				},
+			},
+		},
+		ExpectCreates: []runtime.Object{
+			&kpackbuildv1alpha1.Image{
```
glyn: Inline declarations will end up being unmaintainable. As the tests are fleshed out to cover error percolation and other scenarios, we'll end up with a lot of duplication which will require parallel maintenance.

The main issue with adding helper functions is the potential for their names clashing with other similar functions in the same package. One solution would be to define helper function variables in the scope of the test function. A downside of this is that they need to be defined before being used and so need to appear at the top. Another solution would be to use longer names and a naming convention to avoid clashes. A more radical solution would be to move each controller to a separate package.

PR #250, opened 2019-12-17
Add InMemoryProvider for testing

scothis:
The in-memory provider uses liiklus' memory record store. It does not
use an external broker or any external dependencies. The InMemoryProvider
should not be used for any workloads where reliable delivery is
important.

It is intended only for use in test environments where standing up a
full message broker is undesirable.


ericbottard:
Merged as squashed 37364e6

PR #251, opened 2019-12-17
Retry installing cert-manager with kapp

scothis:
It frequently fails because of a known issue in kapp. Perhaps retrying
is enough to get it to work.

Refs k14s/kapp#12


scothis:
Both the core and streaming cert-manager installs would have failed without this change. With it they both succeeded

PR #252, opened 2019-12-18
Improve handling of Reconcile return values

glyn:
* Use a function to verify the Result and error values (rather than expecting an error message).

* Assert the expected value of the Result.Requeue value. This is done in  addition to the verify function so that the requeue value is checked even when  no verify function is provided.


pkg/controllers/testing/table.go
```diff
@@ -56,21 +57,27 @@ type Testcase struct {
...
+	// ShouldRequeue is true if and only if reconciliation is expected to return a Result with Requeue set to true
+	ShouldRequeue bool
+
+	// Verify provides the reconciliation Result and error for custom assertions
+	Verify VerifyFunc
 
 	// Hooks, if provided, wrap the corresponding client method. They are typically used to inject errors. The object
 	// will be recorded as created if and only if the hook calls the wrapped method.
 	CreateHook CreateHook
 	GetHook    GetHook
```
fbiville: for another PR: it would be helpful to document `{Create,Get}Hook`, I currently do not understand what purpose they serve when looking at their usage in the tests

pkg/controllers/testing/table.go
```diff
@@ -86,17 +93,18 @@ func (r *Testcase) Test(t *testing.T, scheme *runtime.Scheme, factory Factory) {
...
 	c := factory(t, r, clientWrapper, tracker, log)
+
 	// Run the Reconcile we're testing.
-	if _, err := c.Reconcile(reconcile.Request{
+	result, err := c.Reconcile(reconcile.Request{
 		NamespacedName: r.Key,
-	}); (err != nil) != r.ShouldErr {
+	})
+
+	if (err != nil) != r.ShouldErr {
```
fbiville: nit: `r` as a variable name is not very clear

glyn: `r` -> `tc`

pkg/controllers/testing/table.go
glyn: Tried to improve that and added an example. See also the godoc for the `CreateHook` and `GetHook` types.

PR #259, opened 2019-12-20
WIP- testing fats streaming CI flake

ekcasey:
* keda will not scale from zero before there is traffic https://github.com/kedacore/keda/issues/523


scothis:
stale

PR #262, opened 2019-12-23
Use InMemoryProvider for FATS tests

scothis:
Not using kafka will reduce strain on the test cluster and should speed
up the test.


ericbottard:
LGTM, maybe we can rename the provider instance to avoid false cognitive assumptions

fbiville:
Rebased and merged as c0581a5e31b7040a690926fa5a57cebde9394dfc

scothis:
This should not have been merged, as the change is failing CI

scothis:
stale

PR #264, opened 2019-12-24
Switch client hooks for reactors

scothis:
Reactors have full access to the client request and are able to
selectively produce errors or observe/mutate the query.

In addition:
- ExpectUpdate and ExpectDelete are implemented
- ShouldRequeue is renamed to ExpectResult, and is compared to the
  result when the request doesn't err.


pkg/controllers/streaming/processor_controller_test.go
```diff
@@ -285,27 +277,27 @@ func TestReconcile(t *testing.T) {
...
 		ExpectTracks: []rtesting.TrackRequest{
 			rtesting.CreateTrackRequest("", "ConfigMap", "", processorImages).By(testNamespace, testName),
 			rtesting.CreateTrackRequest("build.projectriff.io", "Container", testNamespace, testContainer).By(testNamespace, testName),
 		},
-		GetHook:       getErrorOn(&v1alpha1.Container{}, testNamespace, testContainer, scheme),
-		ShouldErr:     true,
-		ShouldRequeue: true,
-		Verify:        rtesting.AssertErrorEqual(testError),
+		WithReactors: []rtesting.ReactionFunc{
+			rtesting.InduceFailure("get", "Container"),
```
fbiville: it seems that `getErrorOn` was a little more precise than what `InduceFailure` does here.
Is there a way to induce a failure on a Container with a particular name in a particular namespace?

scothis: create a new reactor

PR #266, opened 2019-12-27
Test build controllers

scothis:
Minimal changes to existing controllers. Adding tests for the as implemented behavior.

- create factories to help generate test scaffolds
- test ApplicationReconciler
- test FunctionReconciler
- test ContainerReconciler
  - incomplete due to needing to mock the digest resolution
- test ClusterBuilderReconciler
- test CredentialReconciler


scothis:
Glyn and I agreed to work in master since there is no impact to users or other components. The RFC, in this case, is to reflect our decisions

PR #268, opened 2020-01-02
Add tests for the Knative Deployer controller

scothis:
Also enhances the InjectFailure method to handle more fine grain
failures, including specifying the error.


pkg/controllers/testing/client.go
```diff
@@ -258,11 +260,58 @@ func (w *statusWriterWrapper) Patch(ctx context.Context, obj runtime.Object, pat
 //      // Makes calls to create stream return an error.
 //      rifftesting.InduceFailure("create", "Stream"),
 //   },
-func InduceFailure(verb, resource string) ReactionFunc {
+func InduceFailure(verb, kind string, o ...InduceFailureOpts) ReactionFunc {
+	opts := InduceFailureOpts{}
+	if len(o) > 0 {
```
sbawaska: throw error on more than one InduceFailureOpts?

pkg/controllers/testing/client.go
```diff
@@ -258,11 +260,58 @@ func (w *statusWriterWrapper) Patch(ctx context.Context, obj runtime.Object, pat
...
-		if !action.Matches(verb, resource) {
+		if !action.Matches(verb, kind) {
+			return false, nil, nil
+		}
+		if opts.Namespace != "" && opts.Namespace != action.GetNamespace() {
+			return false, nil, nil
+		}
+		if opts.Name != "" {
+			switch a := action.(type) {
+			case GetAction:
```
sbawaska:
```suggestion
			case GetAction, PatchAction, DeleteAction:
```
multiple conditions can be combined.

pkg/controllers/testing/client.go
```diff
@@ -258,11 +260,58 @@ func (w *statusWriterWrapper) Patch(ctx context.Context, obj runtime.Object, pat
...
+				if opts.Name == a.GetName() {
+					return false, nil, nil
+				}
+			case CreateAction:
```
sbawaska:
```suggestion
			case CreateAction, UpdateAction:
```

pkg/controllers/testing/client.go
```diff
@@ -258,11 +260,58 @@ func (w *statusWriterWrapper) Patch(ctx context.Context, obj runtime.Object, pat
...
+		if !action.Matches(verb, kind) {
+			return false, nil, nil
+		}
+		if opts.Namespace != "" && opts.Namespace != action.GetNamespace() {
+			return false, nil, nil
+		}
+		if opts.Name != "" {
+			switch a := action.(type) {
+			case GetAction:
+				if opts.Name == a.GetName() {
```
sbawaska:
```suggestion
				if opts.Name != a.GetName() {
```
I think all the name checks be `!=`

pkg/controllers/testing/client.go
scothis: doesn't work with typed switches, but I can introduce an interface that abstracts the common bits

PR #270, opened 2020-01-03
Refine status condition usage

scothis:
- create a factory for conditions
- update .StatusConditions() to accept the factory
- use the condition factory in tests


pkg/controllers/testing/factories/condition.go
```diff
@@ -0,0 +1,109 @@
...
+	})
+}
+
+func (f *condition) False() *condition {
+	return f.Mutate(func(c *apis.Condition) {
+		c.Status = corev1.ConditionFalse
+	})
+}
+
+func (f *condition) Reason(reason, message string) *condition {
```
sbawaska: Although `.False().Reason()` reads better when used, Should `False()` set the reason and message to prevent reason not being set? 
I don't feel strongly about this.

scothis: I tried this at one point. `Reason()` isn't always used with `False()`. While closely related, they are distinct and it's better to treat them independently.

PR #274, opened 2020-01-06
Normalize enqueuing tracked resources

scothis:
Centralize the superset of existing capabilities that were present in
each controller.


ericbottard:
LGTM overall.
Couldn't the extracted function become a method of `Tracker` though, thus avoiding passing it around?

scothis:
> Couldn't the extracted function become a method of Tracker

I thought about that, but didn't want to couple the tracker to controller-runtime.

PR #286, opened 2020-01-09
[Spike] Use factories instead of Kubernetes objects in test table rows

glyn:
The general idea is to avoid the proliferation of `Get()` calls and use factories instead of built objects in table rows.

The spike makes the necessary changes to the test framework and factories, but applies the changes to a single test.

### DIFF
--- a/pkg/controllers/testing/factories/objectmeta.go
+++ b/pkg/controllers/testing/factories/objectmeta.go
@@ -22,6 +22,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/projectriff/system/pkg/controllers/testing"
 )
 
 type ObjectMeta interface {
@@ -34,7 +36,7 @@ type ObjectMeta interface {
 	AddLabel(key, value string) ObjectMeta
 	AddAnnotation(key, value string) ObjectMeta
 	Generation(generation int64) ObjectMeta
-	ControlledBy(owner metav1.Object, scheme *runtime.Scheme) ObjectMeta
+	ControlledBy(owner testing.Factory, scheme *runtime.Scheme) ObjectMeta
 	Created(sec int64) ObjectMeta
 	Deleted(sec int64) ObjectMeta
 }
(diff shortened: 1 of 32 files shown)


pkg/controllers/testing/factory.go
```diff
@@ -0,0 +1,25 @@
...
+*/
+
+package testing
+
+import "k8s.io/apimachinery/pkg/runtime"
+
+// Factory creates Kubernetes objects
+type Factory interface {
+	// Get returns a Kubernetes object
+	Get() runtime.Object
```
scothis: perhaps use `apis.Object` since it's the superset of `runtime.Object` and `metav1.Object`

scothis: I'd also suggest leaving `Get()` as the typed struct, and introduce `GetObject()` to get the interface

pkg/controllers/testing/factories/objectmeta.go
```diff
@@ -34,7 +34,7 @@ type ObjectMeta interface {
 	AddLabel(key, value string) ObjectMeta
 	AddAnnotation(key, value string) ObjectMeta
 	Generation(generation int64) ObjectMeta
-	ControlledBy(owner metav1.Object, scheme *runtime.Scheme) ObjectMeta
+	ControlledBy(owner runtime.Object, scheme *runtime.Scheme) ObjectMeta // TODO: should owner be a Factory?
```
scothis: that would be reasonable

pkg/controllers/testing/table.go
```diff
@@ -78,12 +78,22 @@ type VerifyFunc func(t *testing.T, result controllerruntime.Result, err error)
...
 	if tc.Skip {
 		t.SkipNow()
 	}
-	clientWrapper := newClientWrapperWithScheme(scheme, tc.GivenObjects...)
+
+	// Record the given objects
+	givenObjects := make([]runtime.Object, 0, len(tc.GivenObjects))
+	originalGivenObjects := make([]runtime.Object, 0, len(tc.GivenObjects))
+	for _, obj := range tc.GivenObjects {
+		object := obj.Get().DeepCopyObject()
```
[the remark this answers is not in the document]
glyn: Good catch! Fixed.

pkg/controllers/testing/factory.go
glyn: > I'd also suggest leaving `Get()` as the typed struct, and introduce `GetObject()` to get the interface

When would the typed struct be needed rather than the interface?

glyn: > perhaps use `apis.Object` since it's the superset of `runtime.Object` and `metav1.Object`

Yes, very nice.

pkg/controllers/testing/factories/objectmeta.go
glyn: Done.

PR #287, opened 2020-01-09
Improve factory terminology

glyn:
The types in the "factories" package would more commonly be called (fluent)
builders.

The construction method `Get()` doesn't make it obvious whether a new instance
is being created or (unlikely but conceivable) a previously created instance
is being reused. Using builder terminology, a 'Build()' method has no
connotation of reusing something previously built.

### DIFF
--- a/pkg/controllers/build/application_controller_test.go
+++ b/pkg/controllers/build/application_controller_test.go
@@ -31,7 +31,7 @@ import (
 	kpackbuildv1alpha1 "github.com/projectriff/system/pkg/apis/thirdparty/kpack/build/v1alpha1"
 	"github.com/projectriff/system/pkg/controllers/build"
 	rtesting "github.com/projectriff/system/pkg/controllers/testing"
-	"github.com/projectriff/system/pkg/controllers/testing/factories"
+	"github.com/projectriff/system/pkg/controllers/testing/builders"
 	"github.com/projectriff/system/pkg/tracker"
 )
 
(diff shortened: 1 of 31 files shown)


glyn:
> I used `.Build()` in my first draft, but backed away since Function, Application and Container are build resources. I wanted to avoid overloading the term.

I see your point, although it wouldn't bother me personally.

> That said, I'm also not sold on `.Get()`.

I'd be equally happy with `Construct()` or `Create()`. wdyt?

pkg/client/clientset/versioned/typed/build/v1alpha1/application.go
```diff
@@ -63,7 +63,7 @@ func newApplications(c *BuildV1alpha1Client, namespace string) *applications {
 	}
 }
 
-// Get takes name of the application, and returns the corresponding application object, and an error if there is any.
+// Build takes name of the application, and returns the corresponding application object, and an error if there is any.
```
scothis: where did this change come from?

glyn: Oops - I clearly failed to dampen Goland's enthusiasm during one of the refactorings. Fixed.

glyn:
@scothis Let's defer this PR until https://github.com/projectriff/system/pull/286 is implemented fully. then it will be trivial to rename `Build` to whatever we decide on.

PR #291, opened 2020-01-13
Factor out some duplicated code

glyn:
The code to compare expected and actual objects is repetitive and could be factored out. I'm not convinced it is worth the extra complexity, but I'm putting up this spike for discussion.


pkg/controllers/testing/client.go
```diff
@@ -30,13 +30,17 @@ import (
...
+type SingularAction interface {
+	GetObject() runtime.Object
+}
+
 type clientWrapper struct {
 	client              client.Client
 	scheme              *runtime.Scheme
-	createActions       []CreateAction
-	updateActions       []UpdateAction
+	createActions       []SingularAction
```
scothis: the public interface should continue to use the more specific actions types

pkg/controllers/testing/table.go
```diff
@@ -214,27 +183,32 @@ func (tc *Testcase) Test(t *testing.T, scheme *runtime.Scheme, factory Reconcile
...
-			t.Errorf("Missing status update: %#v", exp.Get())
+	compareActions(t, "status update", tc.ExpectStatusUpdates, clientWrapper.statusUpdateActions, statusSubresourceOnly, ignoreLastTransitionTime, safeDeployDiff, cmpopts.EquateEmpty())
+
+	// Validate the given objects are not mutated by reconciliation
+	if diff := cmp.Diff(originalGivenObjects, givenObjects, safeDeployDiff, cmpopts.EquateEmpty()); diff != "" {
+		t.Errorf("Given objects mutated by test %s (-expected, +actual): %v", tc.Name, diff)
+	}
+}
+
+func compareActions(t *testing.T, actionName string, expectedActionFactories []Factory, actualActions []SingularAction, diffOptions ...cmp.Option) {
```
scothis:
```suggestion
func compareActions(t *testing.T, actionName string, expectedActionFactories []Factory, actualActions []objectAction, diffOptions ...cmp.Option) {
```

https://github.com/projectriff/system/blob/873427cb3b598a0159d21e1e934ddb89ef63b5da/pkg/controllers/testing/client.go#L309-L312

pkg/controllers/testing/client.go
```diff
@@ -30,13 +30,17 @@ import (
 	"github.com/projectriff/system/pkg/apis"
 )
 
+type SingularAction interface {
```
scothis: rm, after other suggestions, this is unused

pkg/controllers/testing/client.go
glyn: Not sure about this. Firstly, `createActions` etc. are unexported, so they don't affect the public interface. Secondly, if we leave their types as `CreateAction` etc. then the calls to `compareAction` won't compile, e.g.:
```
Cannot use 'clientWrapper.createActions' (type []CreateAction) as type []objectAction
```
The problem is that even though `CreateAction` is compatible with `objectAction`, the corresponding slices are incompatible. This can perhaps be worked around by making the `compareAction` parameter type `interface{}` and then using reflection inside `compareAction`, but I doubt the result would be better than the current repetitive solution.

So why do the slices in the struct need to use the more specific action types?

glyn: Here's the necessary reflective code. I think the cure is worse than the disease.
```
func compareActions(t *testing.T, actionName string, expectedActionFactories []Factory, actualActions interface{}, diffOptions ...cmp.Option) {
	t.Helper()

	aa := reflect.ValueOf(actualActions)
	if aa.Kind() != reflect.Slice {
		t.Fatal("actualActions must be a slice type")
	}

	for i, exp := range expectedActionFactories {
		if i >= aa.Len() {
			t.Errorf("Missing %s: %#v", actionName, exp.Create())
			continue
		}
		results :=  aa.Index(i).MethodByName("GetObject").Call([]reflect.Value{})
		if len(results) != 1 {
			t.Fatal("GetObject did not return a single value")
		}
		r := results[0]
		if r.IsNil() {
			t.Fatal("GetObject returned nil value")
		}
		actual, ok := r.Interface().(runtime.Object)
		if ! ok {
			t.Fatal("Could not convert result of GetObject to runtime.Object")
		}

		if diff := cmp.Diff(exp.Create(), actual, diffOptions...); diff != "" {
			t.Errorf("Unexpected %s (-expected, +actual): %s", actionName, diff)
		}
	}
	if actual, expected := aa.Len(), len(expectedActionFactories); actual > expected {
		for i := expected; i < actual; i++ {
			t.Errorf("Extra %s: %#v", actionName, aa.Index(i))
		}
	}
}
```

scothis: I mixed up the client with the table. Changing the type is fine.

pkg/controllers/testing/table.go
glyn: Adopted.

pkg/controllers/testing/client.go
glyn: True, nice.

glyn:
@scothis So: do you think this change is worth merging? I don't feel strongly.

glyn:
The change seems consistent with the other reflective code recently introduced. Merging.

PR #292, opened 2020-01-14
[spike] Child reconcilers

scothis:
Abstract common parent-to-child reconciliation logic. This normalizes a basic pattern for reconciliation while still allowing custom logic as needed.

Handled use cases:
- create a child resource
- update a child resource
- remove duplicate child resources
- remove an unneeded child resource
- reflect child on parent's status

Children can either use static or generated names.


glyn:
Although I love removing duplication, I'm nervous that this PR makes the affected controllers harder to read and understand, even for people familiar with kubebuilder-generated controllers. The reflective code is also rather difficult to read (and may be tricky to diagnose, debug, and maintain).

Can the same pattern be applied to all the other controllers or do only some of them fit the pattern? This wasn't immediately obvious. In particular, is the pattern limited to controllers whose _raison d'être_ is to manage child resources?

Perhaps some godoc explaining the design of `ParentReconciler`, `SyncReconciler`, and `ChildReconciler` would help. The fact that these are all in a file called `child.go` is rather distracting.

It's not clear to me why those types are likely to be sufficient for all controllers. I feel like I'm missing a "theory" of reconciliation which explains why the pattern in this PR is sufficient for implementing a general controller. (Compare the table-driven controller tests with their intuitively obvious concepts of objects present before reconciliation, expected creates, status updates, etc.)

If the pattern is not sufficiently general, then I worry that we'll run into situations where it constrains the implementation too much and will result in obscure code to work around the pattern.

Sorry this has been a bit of a negative review. If the above points can be addressed, then it may be easier to "grok" what's going on and appreciate its beauty. ;-)

scothis:
@glyn that's fair. This PR is still an early work in progress. I'm interested in general feedback at this point.

The parent reconciler provides a shell to host sub reconcilers. It looks up the requested resource, makes sure it exists, passes the resource to each sub reconciler in turn and then updates the parent resource's status (if needed).

There are two types of sub resources: sync and child.

The child reconciler is where this work started. For a parent resource create zero-to-one child resources. The relationship between the parent and child resource is managed by the reconciler.

The sync reconciler (not a great name) is an escape hatch to allow for custom behavior to be plugged into a parent. It doesn't provide much functionality directly.

The most complete reconcilers so far are Application and Function. Each has a parent with a sync and child sub reconciler. The sync reconciler is used to resolve the target image with a base value stored in a ConfigMap that is looked up. The child reconciler creates the kpack Image to perform the build.

I agree that using `interface{}` as the function signature will make it harder to know what is expected, but since golang doesn't have generics, there's no other way to provide type specific callbacks. While subjective, needing to do a bit of extra work to define the function signature is nicer than casting each argument inside each method.

These patterns are not one size fits all, nor are they intended to be. The credential and clusterbuilder reconcilers don't fit the parent model since they aggregate multiple resources into a single resource. They can continue to use the controller-runtime reconciler interface directly.

Overall, there is still a lot to refine, but I'm happy with the potential.

scothis:
> I'm nervous that this PR makes the affected controllers harder to read and understand, even for people familiar with kubebuilder-generated controllers.

kubebuilder is a lightly opinionated tool for scaffolding controllers. Those controllers use controller-runtime to provides support up to the Reconcile method. The implementation of the Reconcile method is left as an exercise to the reader as that's your domain logic.

This PR starts at the Reconcile method and paves cow paths within our reconcilers.

glyn:
> > I'm nervous that this PR makes the affected controllers harder to read and understand, even for people familiar with kubebuilder-generated controllers.
> 
> kubebuilder is a lightly opinionated tool for scaffolding controllers. Those controllers use controller-runtime to provides support up to the Reconcile method. The implementation of the Reconcile method is left as an exercise to the reader as that's your domain logic.
> 
> This PR starts at the Reconcile method and paves cow paths within our reconcilers.

Yes, fair enough. What I think will make these Reconcile methods harder to read is the amount of unfamiliar machinery involved. Improved documentation will help, but the abstraction has to feel natural if it's really going to be successful.

scothis:
@glyn please take another look. I've broadened the coverage to complete the core and knative runtimes along with the Application and Function resources. Much of a kludge has been removed, although I'm not quite happy with the logging yet, it's ok.

glyn:
> @glyn please take another look. I've broadened the coverage to complete the core and knative runtimes along with the Application and Function resources. Much of a kludge has been removed, although I'm not quite happy with the logging yet, it's ok.

Looks better, but the changes still don't scratch where I'm itching. How about adding some godoc to the public types in `reconciler.go` which explain when the parent/child pattern is applicable and how it should be used? In particular, I'd like to understand abstractly how to think about a `SubReconciler`, such as how desired and expected state is meant to be split between a parent and its children and how reconciliation occurs with that structure. I know I could probably reverse engineer some of that intention from the code, but I'd rather not. ;-)

PR #294, opened 2020-01-16
Add container concurrency to knative deployer

joshrider:
Adds ContainerConcurrency to the Knative Deployer's DeployerSpec. When set, will
limit the number of concurrent requests that will be processed by a pod at any
one time.

Addresses projectriff/riff#1340


pkg/controllers/knative/deployer_controller.go
```diff
@@ -324,7 +324,8 @@ func (r *DeployerReconciler) constructConfigurationForDeployer(deployer *knative
 					Annotations: annotations,
 				},
 				Spec: servingv1.RevisionSpec{
-					PodSpec: template.Spec,
+					PodSpec:              template.Spec,
+					ContainerConcurrency: r.getConcurrency(deployer),
```
scothis: the helper isn't doing much help

```suggestion
					ContainerConcurrency: deployer.Spec.ContainerConcurrency,
```

joshrider: "helper" removed.

PR #312, opened 2020-02-05
Check for owned conflicting resources

### LINKED ISSUE #310: There is an existing Gateway "test" that the KafkaGateway does not own.
> There is an existing Gateway "test" that the KafkaGateway does not own.

This is expected if there is legitimately another Gateway with the name `test` in the same namespace, however, sometimes this error occurs when the Gateway is owned by KafkaGateway. There is a chance that the desired resource has been created correctly, but then the parent reconciliation fails, when the parent reconciler requeues, the child object has not entered into the informer cache, so the reconciler thinks the resource doesn't exist and attempts to create it, causing the conflict.

When we get a conflict creating a named resource, we should fetch that resource and compare the owner before erring. While the resource will correctly become ready, the error is enough for the cli to catch and report.

Refs https://github.com/projectriff/release/pull/151/checks?check_run_id=426240425#step:9:27

scothis:
When encountering an existing resource that causes a create to fail,
lookup the resource to see if it is the child we wanted to create. This
can happen due to a delay for existing resources getting into the
informer cache.

Changes include:
- adding an APIReader to controller.Config for direct access to the api
  server bypassing the cache
- adding APIGivenObjects to the reconciler table test case to populate a
  test api reader
- detects failed creates due to an existing resource, looks up the
  conflicting resource directly on the api server and skips updating
  the parent's status if the conflict is the child

Resolves #310


pkg/controllers/reconcilers.go
```diff
@@ -323,6 +326,16 @@ func (r *ChildReconciler) Reconcile(ctx context.Context, parent apis.Object) (ct
 	child, err := r.reconcile(ctx, parent)
 	if err != nil {
 		if apierrs.IsAlreadyExists(err) {
+			// check if the failed create is owned by the parent.
+			// the created child from a previous turn may be slow to appear in the informer cache, but shouldn't appear
+			// on the parent as being not ready.
+			apierr := err.(apierrs.APIStatus)
```
glyn: What ensures that this type assertion will not panic?

pkg/controllers/reconcilers.go
```diff
@@ -323,6 +326,16 @@ func (r *ChildReconciler) Reconcile(ctx context.Context, parent apis.Object) (ct
...
+			maybechild := r.ChildType.DeepCopyObject().(apis.Object)
```
glyn:
```suggestion
			maybeChild := r.ChildType.DeepCopyObject().(apis.Object)
```

pkg/controllers/streaming/inmemorygateway_controller_test.go
```diff
@@ -336,16 +336,54 @@ func TestInMemoryGatewayReconciler(t *testing.T) {
...
+		APIGivenObjects: []rtesting.Factory{
+			gatewayGiven,
+		},
+		ShouldErr: true,
+		ExpectTracks: []rtesting.TrackRequest{
+			rtesting.NewTrackRequest(inMemoryGatewayImagesConfigMap, inMemoryGateway, scheme),
+		},
+		ExpectEvents: []rtesting.Event{
+			rtesting.NewEvent(inMemoryGateway, scheme, corev1.EventTypeWarning, "CreationFailed",
+				`Failed to create Gateway "%s":  "%s" already exists`, testName, testName),
```
glyn: Why should the above event be emitted in this case?

If there is a good reason for emitting the above event, wouldn't it be helpful to give more information so that this case can be differentiated from the more serious problem of a non-owned child already existing?

pkg/controllers/reconcilers.go
```diff
@@ -323,6 +326,16 @@ func (r *ChildReconciler) Reconcile(ctx context.Context, parent apis.Object) (ct
...
+			_ = r.APIReader.Get(ctx, types.NamespacedName{Namespace: parent.GetNamespace(), Name: apierr.Status().Details.Name}, maybechild)
```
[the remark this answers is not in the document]
scothis: This is how the controller-runtime clients work. An empty object is passed into the `.Get()` method by reference and the client populates it.

> Since we want to get the object which already existed when trying to create the child, why not use namespace r.ChildType.GetNamespace() and name r.ChildType.GetName() in the call to r.APIReader.Get()?

`ChildType` is empty, there is no namespace or name.

> I'm not sure why a deep copy of the child is used to prime maybechild. Don't we actually want to get the object which conflicted?

We don't want to mutate the `ChildType` which is shared with all turns of this reconciler.

> If the call to r.APIReader.Get() fails, then maybechild may not be updated. Wouldn't it then appear to be controlled by the parent, which would be misleading (if the conflicting, but (temporarily?) inaccessible, object isn't owned by the parent)?

We're already in an error handler that takes precedence. The lookup is an attempt to gracefully handle a corner case. If it fails, we don't detect that corder case and the mainline error handling continues.

pkg/controllers/streaming/inmemorygateway_controller_test.go
scothis: I'm a bit torn by this event. It distracts more than it informs, but it's also also accurate as to what the controller did. Excluding the event introduces a lot more complexity into the controller flow that, at least for now, is worth avoiding.

pkg/controllers/reconcilers.go
scothis: the if check just above it

pkg/controllers/reconcilers.go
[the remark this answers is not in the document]
glyn: Ok, that's helpful. Thanks.

PR #314, opened 2020-02-05
Refactor/document type objects

glyn:
This change is incomplete and is for review in draft form only.

### DIFF
--- a/pkg/controllers/reconcilers.go
+++ b/pkg/controllers/reconcilers.go
@@ -54,14 +54,34 @@ type Config struct {
 	Tracker   tracker.Tracker
 }
 
+// ObjectType represents the type of a Kubernetes object. Fields of type ObjectType are
+// initialized with an empty object and not modified subsequently.
+type ObjectType struct {
+	runtime.Object
+}
+
+// NewObjectType constructs an ObjectType from the given empty object.
+func NewObjectType(obj runtime.Object) ObjectType {
+	return ObjectType{obj}
+}
+
+// asObject returns a deep copy of the object type as an apis.Object.
+func (ot ObjectType) asObject() apis.Object {
+	return ot.DeepCopyObject().(apis.Object)
+}
+
+func (ot ObjectType) typeName() string {
+	return typeName(ot)
+}
+
 // ParentReconciler is a controller-runtime reconciler that reconciles a given
 // existing resource. The ParentType resource is fetched for the reconciler
 // request and passed in turn to each SubReconciler. Finally, the reconciled
 // resource's status is compared with the original status, updating the API
 // server if needed.
 type ParentReconciler struct {
-	// Type of resource to reconcile
-	Type runtime.Object
+	// ObjectType of resource to reconcile
+	Type ObjectType
 
 	// SubReconcilers are called in order for each reconciler request. If a sub
 	// reconciler errs, further sub reconcilers are skipped.
@@ -71,7 +91,7 @@ type ParentReconciler struct {
 }
 
 func (r *ParentReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	bldr := ctrl.NewControllerManagedBy(mgr).For(r.Type)
+	bldr := ctrl.NewControllerManagedBy(mgr).For(r.Type.asObject())
 	for _, reconciler := range r.SubReconcilers {
 		err := reconciler.SetupWithManager(mgr, bldr)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controllers/reconcilers.go
```diff
@@ -67,8 +87,18 @@ type ParentReconciler struct {
 	Config
 }
 
+// NewParentReconciler returns a ParentReconciler given an empty object to represent the parent's type, a Config,
+// and a list of SubReconcilers.
+func NewParentReconciler(objType runtime.Object, cfg Config, subReconcilers ...SubReconciler) *ParentReconciler {
```
scothis: I'm 👎 on constructors that mask the semantic value of field names for positional args without adding meaningful behavior.

glyn: Ok, I've inlined the constructor for now. We can always introduce a constructor or builder later if necessary.

pkg/controllers/reconcilers.go
```diff
@@ -54,14 +54,34 @@ type Config struct {
...
+
 // ParentReconciler is a controller-runtime reconciler that reconciles a given
 // existing resource. The ParentType resource is fetched for the reconciler
 // request and passed in turn to each SubReconciler. Finally, the reconciled
 // resource's status is compared with the original status, updating the API
 // server if needed.
 type ParentReconciler struct {
-	// Type of resource to reconcile
-	Type runtime.Object
+	// ObjectType of resource to reconcile
```
scothis: godoc should be use the field name, not the type

pkg/controllers/reconcilers.go
```diff
@@ -54,14 +54,34 @@ type Config struct {
...
+}
+
+// NewObjectType constructs an ObjectType from the given empty object.
+func NewObjectType(obj runtime.Object) ObjectType {
+	return ObjectType{obj}
+}
+
+// asObject returns a deep copy of the object type as an apis.Object.
+func (ot ObjectType) asObject() apis.Object {
+	return ot.DeepCopyObject().(apis.Object)
```
scothis: by convention, methods do not return copies. It's the responsibility of the caller to create a copy before inflicting mutations that should be localized.

pkg/controllers/reconcilers.go
```diff
@@ -54,14 +54,34 @@ type Config struct {
...
+	runtime.Object
+}
+
+// NewObjectType constructs an ObjectType from the given empty object.
+func NewObjectType(obj runtime.Object) ObjectType {
+	return ObjectType{obj}
+}
+
+// asObject returns a deep copy of the object type as an apis.Object.
+func (ot ObjectType) asObject() apis.Object {
```
scothis: should be exported, or drop the method and lean on the current `Object` field

```suggestion
func (ot ObjectType) Object() apis.Object {
```

pkg/controllers/reconcilers.go
```diff
@@ -54,14 +54,34 @@ type Config struct {
 	Tracker   tracker.Tracker
 }
 
+// ObjectType represents the type of a Kubernetes object. Fields of type ObjectType are
+// initialized with an empty object and not modified subsequently.
+type ObjectType struct {
+	runtime.Object
```
scothis: this shouldn't be exported

pkg/controllers/reconcilers.go
```diff
@@ -54,14 +54,34 @@ type Config struct {
...
+}
+
+// NewObjectType constructs an ObjectType from the given empty object.
+func NewObjectType(obj runtime.Object) ObjectType {
+	return ObjectType{obj}
```
scothis: use named fields

pkg/controllers/reconcilers.go
```diff
@@ -54,14 +54,34 @@ type Config struct {
...
+}
+
+// asObject returns a deep copy of the object type as an apis.Object.
+func (ot ObjectType) asObject() apis.Object {
+	return ot.DeepCopyObject().(apis.Object)
+}
+
+func (ot ObjectType) typeName() string {
```
scothis: should be exported

```suggestion
func (ot ObjectType) Name() string {
```

glyn:
> Even with my inline suggestions, I think we should abandon this PR:
> 
> 1. controller-runtime uses bare `runtime.Object`s to denote type information. We should strive to use consistent idioms when possible
> 2. ObjectType doesn't simplify our abstract reconcilers
> 3. ObjectType inflicts reconciler implementors with one more thing they need to know (granted it's small)

I'm comfortable with that. Closing.

PR #359, opened 2020-05-04
Consume reconciler runtime

scothis:
Reconciler runtime is a new module extracted from riff system. No new functionality or behavior is included, this is just a factoring of the existing code. By splitting the reconciler framework out, we can show that it is independent of riff, using riff system as an exemplar consumer. If successful, the framework will likely move outside of projectriff.


scothis:
The test failure is a known issue on master

PR #360, opened 2020-05-04
Switch ko to strict mode

scothis:
This will soon be the default, it requires each image path be prefixed
with `ko://`.


scothis:
The error exists on master and is well after the code in this PR run.

